PR #2, opened 2019-05-10
rename to Solus.jl, add NEKI

simonbyrne:
Roughly based on code from @sandreza and @glwagner.

Still needs
- [ ] a higher-level `neki` function?
- [ ] make it work for priors with means
- [ ] some tests: do you have a simple example we can try this on?
- [ ] `eki`: what is the difference between this and neki?


src/Solus.jl
```diff
@@ -0,0 +1,71 @@
...
+    The observed data.
+    """
+    observation::O
+
+    """
+    The space on which the forward model output and observation exist. `DefaultSpace()` is the default.
+    """
+    space::S
+end
+SolusModel(prior, forwardmodel, observation) = SolusModel(prior, forwardmodel, observation, DefaultSpace())
```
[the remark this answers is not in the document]
simonbyrne: woops, missed that one

src/Solus.jl
```diff
@@ -0,0 +1,71 @@
...
+
+    """
+    A forward model maps the parameter ``θ`` to a predicted observation.
+    """
+    forwardmodel::M
+    
+    """
+    The observed data.
+    """
+    observation::O
```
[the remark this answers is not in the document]
simonbyrne: how about `obs`. Then we don't have to argue whether there is 1 or multiple observations.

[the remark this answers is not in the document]
glwagner: like it

src/Solus.jl
```diff
@@ -0,0 +1,83 @@
...
+    """
+    outputs::Vector{O}
+end
+
+"""
+    neki_iter(prob::SolusProblem, ens::Ensemble)
+
+Peform an iteration of NEKI, returning a new `Ensemble` object.
+"""
+function neki_iter(prob::SolusProblem, ens::Ensemble)
```
tapios: Alfredo et al. call this the ensemble Kalman flow (though EKF invites confusion with the filter).

src/Solus.jl
```diff
@@ -41,10 +41,20 @@ SolusProblem(prior, forwardmodel, obs) = SolusProblem(prior, forwardmodel, obs,
...
+
+$(DocStringExtensions.FIELDS)
 """
 struct Ensemble{T,O}
-    inputs::Matrix{T} # on your theta space: each row is a vector of θ
-    outputs::Vector{O} # computed from forward model
+    """
+    A matrix of inputs: each column corresponds to a vector of ``θ``s
+    """
+    inputs::Matrix{T}
```
glwagner: I would prefer that the `inputs` be of the type `Vector{P}`, where `P` is the type of the parameters (possibly itself a `Vector`).

This design means that parameter vectors can be things like `FieldVectors` or other kinds of labeled array-like objects. Labelled arrays make interfacing with some forward model that allows setting parameters values with keyword-argument constructors and `(parameter, value)` pairs simple --- obviating the need that would otherwise exist to map the indices of parameters onto their names for each case specifically. In other words, a user can define a new list of free parameters via

```julia
struct InterestingParameters{T} <: FieldVector{2, T}
    C1 :: T
    C2 :: T
end
```

and pass these to the optimization routine, which will then optimize just the `FreeParameters` `C1` and `C2` of the forward model (of a large number of possible parameters that may be adjusted). In general, it's nice to have flexibility in defining the parameter object. I also think it is a natural abstraction for an ensemble. 

I can provide a use-case if it is desired.

We can provide convenience functions to map between matrices and vectors of vectors for analysis/interfacing if this is needed for some users.

simonbyrne: I did think about that: I was mainly hesitant because we restrict the prior to being MvNormal, and there are a couple of places where we can easily express what we're doing in terms of matrix operations. But I can certainly see the appeal, so lets discuss next week.

tapios: May be worth revisiting what @charleskawczynski did in [#186](https://github.com/climate-machine/CLIMA/pull/186). The parameter structures he defined there ultimately need to be consistent with how we handle parameters here.

glwagner: I see @simonbyrne.

In the case that the ensemble is `ens <: Vector{FieldVector{N, T}}`, I believe we can [use `reinterpret`](https://juliaarrays.github.io/StaticArrays.jl/stable/pages/api.html#Arrays-of-static-arrays-1) to obtain a matrix-like object via `matrix_ens = reinterpret(T, ens)` without allocating memory --- though perhaps that would be the transpose of the object you are thinking about?

PR #7, opened 2019-06-25
GPRWrap

dburov190:


dburov190:
The build fails because somehow `ScikitLearn.jl` can't find Python `sklearn` module through `PyCall` ¯\_(ツ)_/¯

@simonbyrne Could you take a look, please? thanks!

src/GPRWrap/GPRWrap.jl
```diff
@@ -0,0 +1,148 @@
...
+
+sklearn.@sk_import gaussian_process : GaussianProcessRegressor
+sklearn.@sk_import gaussian_process.kernels : (RBF, Matern, WhiteKernel)
+
+
+@with_kw mutable struct GPRWrap
+  """
+  A simple struct to handle Gaussian Process Regression related stuff
+
+  """
```
simonbyrne: Julia docstrings go before the function/type

dburov190: whoops, thanks! I'll change that

PR #12, opened 2020-01-08
Fix package name

charleskawczynski:
Hi @odunbar, @dburov190, and @agarbuno

We need to standardize the package name. Please check that these changes don't break any results/tests. If they don't, feel free to merge and we can push forward. Once this is configured, we can set up some automated CI, so that changes can be made more easily.

Note that I've also changed the tolerance of `1e-8` to `sqrt(eps(Float64))`


docs/src/index.md
```diff
@@ -0,0 +1,3 @@
+# CES.jl
+
+Stochastic Optimization, Learning, Uncertainty and Sampling
```
tapios:
```suggestion
Calibrate, Emulate, Sample
```

src/neki.jl
```diff
@@ -1,28 +1,28 @@
 """
-    neki_iter(prob::SolusProblem, θs, fθs)
+    neki_iter(prob::CESProblem, θs, fθs)
```
tapios:
```suggestion
    eki_iter(prob::CESProblem, θs, fθs)
```
Change `neki` -> `eki` throughout, including in file name.

src/neki.jl
```diff
@@ -1,28 +1,28 @@
...
 
-Peform an iteration of NEKI, returning a new set of proposed `θs`.
+Perform an iteration of NEKI, returning a new set of proposed `θs`.
```
tapios:
```suggestion
Perform an iteration of EKI, returning a new set of proposed `θs`.
```

src/neki.jl
```diff
@@ -1,28 +1,28 @@
...
 
 See eqs. (5.4a-b) from https://arxiv.org/abs/1903.08866.
 """
-function neki_iter(prob::SolusProblem, θs, fθs)
+function neki_iter(prob::CESProblem, θs, fθs)
```
tapios:
```suggestion
function eki_iter(prob::CESProblem, θs, fθs)
```

src/neki.jl
```diff
@@ -1,28 +1,28 @@
...
     covθ = cov(θs)
     meanθ = mean(θs)
-    
+
     m = mean(fθs)
 
     J = length(θs)
     CG = [dot(fθk - m, fθj - prob.obs, prob.space)/J for fθj in fθs, fθk in fθs]
 
     Δt = 1.0 / (norm(CG) + 1e-8)
```
tapios:
```suggestion
    Δt = 1.0 / (norm(CG) +eps)
```

src/neki.jl
```diff
@@ -1,28 +1,28 @@
 """
-    neki_iter(prob::SolusProblem, θs, fθs)
+    neki_iter(prob::CESProblem, θs, fθs)
 
-Peform an iteration of NEKI, returning a new set of proposed `θs`.
+Perform an iteration of NEKI, returning a new set of proposed `θs`.
 
 See eqs. (5.4a-b) from https://arxiv.org/abs/1903.08866.
```
tapios: Update reference, @agarbuno?

test/neki.jl
```diff
@@ -16,14 +16,14 @@ using LinearAlgebra, Distributions
...
-    prob = Solus.SolusProblem(prior, f, y_obs, Solus.CovarianceSpace(Γ))
+    prob = CES.CESProblem(prior, f, y_obs, CES.CovarianceSpace(Γ))
 
     J = 50
     θs = [rand(2) for i in 1:J]
 
     for i = 1:20
         fθs = map(f, θs)
-        θs = Solus.neki_iter(prob, θs, fθs)
+        θs = CES.neki_iter(prob, θs, fθs)
```
tapios:
```suggestion
        θs = CES.eks_iter(prob, θs, fθs)
```

src/neki.jl
charleskawczynski: Since this may change the results, I'm going to leave this for a later PR and open an issue.

ali-ramadhan:
Not sure if the plan is to register this as a Julia package but the name CES.jl might get rejected as it goes against naming guidelines 1 and 4: https://julialang.github.io/Pkg.jl/v1/creating-packages/#Package-naming-guidelines-1

1. Avoid jargon. In particular, avoid acronyms unless there is minimal possibility of confusion.
4. Err on the side of clarity, even if clarity seems long-winded to you.

charleskawczynski:
Thanks @ali-ramadhan! Good point. We'll want to change the name, but I'm not sure what the final name should be. I'm probably not the person to decide on the name. This PR just makes the naming consistent because, before, `CES` and `Solus` were both used but in different places, and I couldn't configure any tests.

charleskawczynski:
Is `CalibrateEmulateSample.jl` too verbose? @agarbuno @dburov190 @odunbar

tapios:
‘CalibrateEmulateSample.JPL’ is good. You can also use ‘EnsembleKalmanSampler’ for ‘EKS’.

tapios:
Please also change the 1e-8 to ‘sqrt(eps)’. Whatever small effect it has on results is not material. We should address such issues when we can. Github issues seem to linger for a long time before being acted upon.

charleskawczynski:
Closes #13

examples/GPR/main.jl
```diff
@@ -4,9 +4,9 @@ import NPZ
 import PyPlot
 const plt = PyPlot
 
-include("../../src/GPR.jl")
+include(joinpath("..","..","src","GPR.jl"))
 
-const gpr_data = NPZ.npzread("../../test/GPR/data/data_points.npy")
+const gpr_data = NPZ.npzread(joinpath("..","..","test","GPR","data","data_points.npy"))
```
dburov190: Maybe a minor point, but is this standard or something? don't see why the proposed way is clearer/better

charleskawczynski: It's more OS friendly. For poor windows users (like myself) this translates to `..\\..\\test\\GPR\\data\\data_points.npy`.

src/neki.jl
charleskawczynski: Should we change this to `https://arxiv.org/pdf/2001.03689.pdf`?

examples/GPR/main.jl
dburov190: oh, I see
funnily enough, I'm myself a windows user (at the time being, hopefully not for long) and this works for me, i.e. I wrote this code :D I guess, mingw makes the difference? anyway, I'm always in for cross-platform

charleskawczynski:
> Seems like this does break a bit of my stuff (examples directory), hold off a little bit, I'll fix that soon and then we can merge.

Let me know if there's anything I can help with.

charleskawczynski:
@dburov190 @odunbar any updates on merging this?

PR #22, opened 2020-01-31
Add Cloudy example to test

charleskawczynski:
- Removes "_mb" from filenames
 - Removes Cloudy dependence from CES.jl by passing objects/emthods as arguments from test

I had to remove windows 32-bit build since it was timing out. Opened #25 to track this.


charleskawczynski:
Thanks for validating the results @bielim!

PR #23, opened 2020-01-31
Update CES code to work with latest Cloudy code

bielim:
Essentially replaced get_src_coalescence with get_int_coalescence

### DIFF
--- a/src/EKI_mb.jl
+++ b/src/EKI_mb.jl
@@ -159,7 +159,7 @@ function run_cloudy(params::Array{Float64, 1}, kernel::KernelTensor{Float64},
     end
 
     # Set up ODE problem: dM/dt = f(M,p,t)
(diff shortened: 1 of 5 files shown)


charleskawczynski:
I may try to just incorporate this new file into #22

charleskawczynski:
Oh, the `GModel.jl` is nearly identical to `EKI.jl`. I think I just needed the new name of the function in Cloudy.

bielim:
Please ignore GModel for now, I should have stashed that instead of committing it (sorry, github newbie here :-)) - GModel will contain everything Cloudy-related, so that e.g. EKI.jl becomes Cloudy-free, but as you can see easily at this point it is basically a copy of EKI.jl

charleskawczynski:
Ok, sounds good, thanks!. I'll close this and continue with #22.

PR #24, opened 2020-02-02
Update README.md

bielim:
An initial description of what CalibrateEmulateSample.jl does - to be extended and modified as the code base develops.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,3 +23,34 @@
 [bors-img]: https://bors.tech/images/badge_small.svg
 [bors-url]: https://app.bors.tech/repositories/11528
 
+
+`CalibrateEmulateSample.jl` solves parameter estimation problems using (approximate) Bayesian inversion. It is designed for problems that require running a computer model that is expensive to evaluate, but can also be used for simple models.
+
+The computer model is supplied by the user – it is a forward model, i.e., it takes certain parameters and produces data that can then be compared with the actual observations. We can think of that model as a parameter-to-data map <a href="https://www.codecogs.com/eqnedit.php?latex=G(\theta):&space;\mathbb{R}^p&space;\rightarrow&space;\mathbb{R}^d" target="_blank"><img src="https://latex.codecogs.com/gif.latex?G(\theta):&space;\mathbb{R}^p&space;\rightarrow&space;\mathbb{R}^d" title="G(\theta): \mathbb{R}^p \rightarrow \mathbb{R}^d" /></a>. For example, <a href="https://www.codecogs.com/eqnedit.php?latex=G" target="_blank"><img src="https://latex.codecogs.com/gif.latex?G" title="G" /></a> could be a global climate model or a model that predicts the motion of a robot arm. 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,3 +23,34 @@
 [bors-img]: https://bors.tech/images/badge_small.svg
 [bors-url]: https://app.bors.tech/repositories/11528
 
+
+`CalibrateEmulateSample.jl` solves parameter estimation problems using (approximate) Bayesian inversion. It is designed for problems that require running a computer model that is expensive to evaluate, but can also be used for simple models.
+
+The computer model is supplied by the user – it is a forward model, i.e., it takes certain parameters and produces data that can then be compared with the actual observations. We can think of that model as a parameter-to-data map <a href="https://www.codecogs.com/eqnedit.php?latex=G(\theta):&space;\mathbb{R}^p&space;\rightarrow&space;\mathbb{R}^d" target="_blank"><img src="https://latex.codecogs.com/gif.latex?G(\theta):&space;\mathbb{R}^p&space;\rightarrow&space;\mathbb{R}^d" title="G(\theta): \mathbb{R}^p \rightarrow \mathbb{R}^d" /></a>. For example, <a href="https://www.codecogs.com/eqnedit.php?latex=G" target="_blank"><img src="https://latex.codecogs.com/gif.latex?G" title="G" /></a> could be a global climate model or a model that predicts the motion of a robot arm.
```
dburov190: I think I'd rather have these as `$G(\theta) \colon \R^p \to \R^d$` etc. because it's almost impossible to read it as plain text, whereas the formulae are simple enough to read it without rendering.

charleskawczynski: Also, less important but, I think I read somewhere that double back tics are preferred to $’s

charleskawczynski: Also, you can use hyper-refs using the format

`[text to hyper-reference](URL)`

Which is nice and compact

bielim: @dburov190: I agree that this is impossible to read as plain text, but at least the rendered version looks nice. Yes, for now the equations are simple enough to be read as plain text, but a) it looks ugly and b) the README may include more complicated equations at some point. So, for me it isn't clear which of the two options is the lesser evil, but I'll change to plain text if that's what people prefer.
Also, why isn't there an easy way to render math in github README pages?

bielim: @charleskawczynski: Thanks, I'll incorporate the double back tics and your suggested hyper-refs format!

bielim: It seems that rendering math in README pages is currently not possible (@charleskawczynski, thanks for exploring this!). We decided to move the content of this README to the docs, where math can be rendered - otherwise, we are either left with non-rendered math equations or with rendered math equations that are based on generating embedded links with a TeX editor that then embeds a gif in GitHub, neither of which is satisfactory.

charleskawczynski:
Also, note that I've changed `ConvenienceFunctions` to `Utilities`

README.md
dburov190: Yep, there's no way to use MathJax or something else to render Latex on github. I think there was some sort of discussion about security reasons.

Whatever, moving math & other details into docs is probably the best decision; Readme's are supposed to be very succinct in general, merely stating the purpose and providing links to documentation and examples.

I'm not sure about the backticks though, never heard of it tbh...

odunbar: I agree with Dima, this would be better suited to a doc; there we can well define the problem, and how the user can interact with the software. 

I feel the Readme would be better to say: this is what you use the software for, this is how you install it and this is where the Docs and examples are.

PR #30, opened 2020-02-06
Add EarthMoversDistance in Julia

charleskawczynski:


charleskawczynski:
Opened [#159](https://github.com/JuliaStats/Distances.jl/pull/159) since this is a better upstream PR.

PR #32, opened 2020-02-13
Separate cloudy and improve CES

bielim:
Revise CES code.
Changes include the following:

- Created a separate module GModel that contains all the functionality  to run the forward model G
--> anything Cloudy-related is contained in   that module. The idea is that this module (and specifically the  function run_G) is all the user has to modify for a specific CES  application (at least for applications where G is a simple model that  can be run without using a cluster etc.)
- The module "Observations.jl" replaces "Truth.jl" - instead of a Truth  object, there is now an "Obs" object that contains the observations. In the general case, CES will deal with observations coming from some external observing system rather than with artificial, model-generated "observations" - in the latter case, the observation can be thought of as the "truth" (since they result from running the model with the  known "true" parameters), but more generally the name "observations" is probably more appropriate.
- All "helper functions" (including some new ones) have been moved to Utilities.jl
- Documented the main structs (EKIObj, GPObj, MCMCObj, Obs)
- Made the covariance (kernel) an (optional) input argument to the GP emulator.


tapios:
This is looking really good! Thanks, @bielim, @charleskawczynski and all who have worked on it.

src/MCMC.jl
```diff
@@ -29,53 +20,83 @@ export find_mcmc_step!
...
+    obs_sample::Vector{FT}
+    obs_cov::Array{FT, 2}
+    obs_covinv::Array{FT, 2}
     prior::Array
     step::Array{FT}
     burnin::I
     param::Vector{FT}
     posterior::Array{FT, 2}
-    log_posterior::Array{Union{FT,Nothing}}
+    log_posterior::Array{Union{FT, Nothing}}
```
charleskawczynski: Can we change this to
```
    log_posterior::Union{Nothing, Array{FT,N}}
```
Also, do we know the rank of `log_posterior`?

src/MCMC.jl
```diff
@@ -29,53 +20,83 @@ export find_mcmc_step!
...
     iter::Array{I}
```
charleskawczynski:
```suggestion
    iter::Array{I,1}
```
?

src/MCMC.jl
```diff
@@ -29,53 +20,83 @@ export find_mcmc_step!
...
     accept::Array{I}
```
charleskawczynski:
```suggestion
    accept::Array{I,1}
```
?

charleskawczynski:
Also, as a general note, I think it's best we change the convention of using `I` for integer to `IT` so that it doesn't clash with `LinearAlgebra`'s identity matrix `I`. But we can change this in another PR.

bielim:
Added updated test/Cloudy/runtests.jl such that the tests run with the latest CES code. @charleskawczynski: GModel.jl currently imports Cloudy modules --> I'm hoping that your code beautification skills can make this go away :-)

charleskawczynski:
> But does this mean that your Jupyter notebook will not run any more?

I'd be fine with removing the Jupyter notebooks altogether and replacing them with [Literate.jl](https://github.com/fredrikekre/Literate.jl) examples in the docs :)

bielim:
@odunbar: The changes I made (GModel.jl, Observations.jl, etc.) are for the general CES code, but in an additional step I also updated test/Cloudy/runtests.jl, which tests the whole CES pipeline on a Cloudy application. That test is basically the Cloudy jupyter notebook I uploaded, just in the form of an automated test instead of a notebook, so in a way the notebook is now redundant. Because of that, I haven't updated the notebook yet - I had a feeling that @charleskawczynski wouldn't be opposed to removing it altogether :-) The argument one can make for keeping it is that some users like the interactive interface of notebooks as a friendly way of being introduced to a topic/code, but the disadvantage of course is that it would be an additional thing to keep in sync with the evolving CES and Cloudy code base. I have no strong opinion on this and am fine with deleting the notebook, but would also be willing to make an effort at keeping it up to date.

bielim:
> But does this mean that your Jupyter notebook will not run any more?
And just to give a clear answer to that question: At the moment the notebook is not running, but if we decide that we want to keep it in the CES repo, I will go ahead and update it to make it run again.

bielim:
> Do you still use the Zscore? I can't see it - therefore is it worth keeping?
> 
> In this vain, i suggest perhaps we always transform inputs to the GP training and prediction. We find the mean and covariance of the input parameters, then do a "Z score":
> 
> new_inputs= input_cov^{-1/2}*(inputs - input_mean)
> 
> Train the GP on this, and apply this to the inputs to the prediction function.
> 
> It makes the GP hyperparameters which users may wish to set into more independent of the problem (e.g they become scaled (approximately)correctly with respect to the geometry of the training points (which in CES is quite important as it can change a lot)).
> 
> It will require storing the inputs_mean and sqrt(inv(cov)) in the GP struct, but these are small.

@odunbar: I will implement this on a separate feature branch!

PR #40, opened 2020-02-18
Added Julia version for EKS

agarbuno:
EKSObj ready for CES.jl . It's basically the same structure as of EKIObj. The main differences are: 
- Prior related objects (mean and covariance).
- Update rule for the ensemble.  
Also, the python blueprint can be found as a reference as well.

### DIFF
--- a/src/EKS-bkp.jl
+++ b/src/EKS-bkp.jl
@@ -1,0 +1,34 @@
+"""
+    eks_iter(prob::CESProblem, θs, fθs)
+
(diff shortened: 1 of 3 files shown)


charleskawczynski:
@agarbuno would you mind if we work off of the main repo? And not the forks? I think it's a bit easier. I can copy this to a new PR if you'd like.

charleskawczynski:
Thanks @agarbuno! Copied this to #41 to avoid working with forks.

PR #42, opened 2020-02-20
Normalize GP input

bielim:
Normalize inputs to the GP training and prediction. 

Normalization is done by centering the inputs (i.e., subtracting their mean) and multiplying them by the square root of the inverse of the input covariance.
Whether or not the GP is trained on normalized inputs can be specified with an optional input argument to GPObj ("normalized"), which defaults to true. If the GP has been trained on normalized
inputs, the "predict" function automatically applies the same same normalization when predicting on new inputs.

The idea is that normalization will make the GP hyperparameters more independent of the problem - e.g., the length scales used for the default kernels can be assumed to be reasonable defaults for many problems.

### DIFF
--- a/src/GPEmulator.jl
+++ b/src/GPEmulator.jl
@@ -50,19 +50,26 @@ models.
 # Fields
 $(DocStringExtensions.FIELDS)
 """
-struct GPObj{FT<:AbstractFloat,GPM}
-    "training points/parameters; N_parameters x N_samples"
+struct GPObj{FT<:AbstractFloat, GPM}
+    "training inputs/parameters; N_samples x N_parameters"
     inputs::Array{FT, 2}
-    "output data (`outputs = G(inputs)`); N_parameters x N_samples"
+    "training data/targets; N_samples x N_data"
     data::Array{FT, 2}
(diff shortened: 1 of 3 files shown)


ali-ramadhan:
> Normalization is done by centering the inputs (i.e., subtracting their mean) and multiplying them by the square root of the inverse of the input covariance.

Sorry if this is a stupid question but would this work no matter the units of the input or do you have to pick suitable units (potentially non-dimensional)?

Like if the inputs were near-surface temperature profiles in °C would it be bad that tropical data is always very positive while Arctic data can have negative numbers? I guess maybe you want your input data to have units of Kelvin or use potential temperature instead maybe.

odunbar:
@ali-ramadhan. The transformation is applied internally just as a means to aid the interpretation of parameters in the Gaussian Process. You don't have to do anything to your inputs (or units) before giving them as inputs to the GPObj, or the Predict function.

So if you give it a training pair of (3C ,100mm), then when you want to predict at 3C it will still give you answer 100mm. Internally however it transforms the 3C -> X and then trains on (X,100mm)

ali-ramadhan:
Thank you for the clarification @odunbar!

PR #44, opened 2020-03-03
Update Cloudy test and Cloudy notebook

bielim:
Update Cloudy test to use lognormal priors.
Update Cloudy notebook such that it can be run with the current version of the CES code. @charleskawczynski: I know that you hate that notebook (or notebooks in general), but until we decide to delete it I'll keep updating it from time to time :-)


charleskawczynski:
>  @charleskawczynski: I know that you hate that notebook (or notebooks in general), but until we decide to delete it I'll keep updating it from time to time :-)

I'm fine with keeping the notebook around if you find that it has value. It may be more efficient to use [Literate](https://github.com/fredrikekre/Literate.jl) to generate notebooks from the existing Julia script though, so that you don't need to maintain duplicate versions.

bielim:
I will look into that, thanks!

PR #45, opened 2020-03-04
Replace FreeParameters.jl with Priors.jl

bielim:
As discussed with @simonbyrne, this branch implements a module `Priors.jl`, whose intent is to replace `FreeParameters.jl`. 
The idea is very simple: All we need is a structure that contains the information on the priors and the corresponding names of the parameters on which we do UQ. The priors are defined directly as Julia Distributions (e.g., `Gamma(5.0, 3.0)`), and so for everything one would typically want to do with these priors (sample, evaluate at specific parameter values, return the parameters, etc.) we can use the already existing functionality provided by Julia's `Distributions.jl` package. For use in CliMA, all priors and parameters should be defined in a separate `.jl` file, as follows:
```
using CalibrateEmulateSample.Priors

# Fall speed parameter for cloud ice
atm_vz_ice = Prior(Uniform(350.0, 1400.0), "atm_vz_ice")

# Max relative humidity threshold for ice cloud
atm_rhmax_ice = Prior(Gamma(2.3, 0.2), "atm_rhmax_ice")

etc.
```
Once specific realizations of random variables have been drawn from the prior distributions (as is done e.g. to construct the initial ensemble in the EKI), it is not necessary to give them any kind of special "free parameter treatment", they are just regular numerical values. This is the key realization underlying this implementation and motivating the replacement of `FreeParameters.jl`.

`Priors.jl` also contains an implementation of a "deterministic distribution" (which is currently not contained in `Distributions.jl`) - this is a "distribution" over only one value (which it takes with probability 1) and can be used e.g. for experiments where parameters need to be treated as fixed values. 

The new Prior structs have already been integrated into the CES code of this branch.


src/MCMC.jl
```diff
@@ -173,24 +172,16 @@ end
...
+    priors = [mcmc.prior[i].dist for i in 1:length(mcmc.prior)]
+    for (idx, prior) in enumerate(priors)
         variances[idx] = var(prior)
     end
 
     if mcmc.algtype == "rwm"
-        #prop_dist = MvNormal(mcmc.posterior[1 + mcmc.iter[1], :], (mcmc.step[1]) * Diagonal(variances))
         prop_dist = MvNormal(zeros(length(mcmc.param)), (mcmc.step[1]^2) * Diagonal(variances))
     end
     sample = mcmc.posterior[1 + mcmc.iter[1], :] .+ rand(prop_dist)
```
odunbar: Did you want to put in some kind of inbounds check? perhaps with just a message if the proposal leaves the support of the prior - something like - " MCMC proposal is leaving the support of the prior for parameter i, consider using a transform or a prior with wider support" ? .

remark on this is that I have switch to using Normal priors under a pushforward through a sigmoid function as opposed to using a uniform prior etc. 

Then resample until it is within bounds.

test/Cloudy/runtests.jl
```diff
@@ -161,7 +162,7 @@ white = Noise(log(2.0))
 # construct kernel
 GPkernel =  kern1 + kern2 + white
 # Get training points    
-u_tp, g_tp = Utilities.extract_GP_tp(ekiobj, N_iter)
+u_tp, g_tp = Utilities.extract_GP_tp(ekiobj, N_iter-1)
```
odunbar: Was this just a typo from before?

odunbar:
Excellent thanks Melanie - this looks exactly as we discussed! I have left a comment or two.
 
Do you think we should do anything more general for the Deterministic prior? like an option to specify an array? (On the other hand I feel if we are doing parameters as "key = value" in some file, then we will not need this functionality)

Otherwise looks clean to me!

src/MCMC.jl
bielim: @odunbar: Sounds like a reasonable thing to do - I'll implement this!

test/Cloudy/runtests.jl
bielim: The second argument to extract_GP_tp just defines the number of EKI layers/iterations to train on - if it's set to N_iter, the initial ensemble is included in the training points, and if it's set to N_iter-1, the initial ensemble is not included in the training points. There is no "right" or "wrong" number of iterations to use though, so either is fine.

charleskawczynski:
> For use in CliMA, all priors and parameters should be defined in a separate .jl file, as follows...

Where is this code snippet supposed to live? in `CalibrateEmulateSample.jl`?

bielim:
Closing this PR, as it has been superseded by #55 (which introduces essentially the same changes as this PR).

PR #46, opened 2020-03-24
Try to fix build

charleskawczynski:


charleskawczynski:
@bielim these are kind of drastic measures (I've pulled out nearly everything in the source code), but in doing so I realized that there are no unit tests for some of the modules, and it would be good to write some. If we merge this, perhaps you can write unit tests for the following modules (in order, since, for example MCMC depends on GPEmulator) and incrementally add things back in
 
 - GPEmulator
 - MCMC
 - Observations
 - GModel
 - Utilities
 - spaces
 - problems
 - GPR

charleskawczynski:
Superceeded by #48

PR #48, opened 2020-04-01
Switch to Github Action docker-based CI

charleskawczynski:
Since we've been having build/environment issues, I thought it might be worth giving a docker-based CI a try. The way this works is:

 - CI is now controlled by a Github Action
 - The GHA checks out the PR/master and builds a docker image on top of the Julia docker image
 - The docker image is then used to run the tests

The hope is that this will stabilize the build environment since everything is built from docker, which should be a clean slate each build, rather than runners whose environments potentially change (which was the case for the last failure on Travis).

This also removes a handful of deps and source code, but we'll add these back in incrementally.


charleskawczynski:
I used the solution [mentioned here](https://github.com/bors-ng/bors-ng/issues/730) to make bors work with GHA

charleskawczynski:
bors try

Issue #56, opened 2020-06-25
Timestep for EKI


odunbar:
https://github.com/CliMA/CalibrateEmulateSample.jl/blob/5fb71c2109b7ff818ceb8339f3351f654234aecc/src/EKI.jl#L126

To aid in preventing the collapse of the ensemble, we should introduce a timestep into the EKI. This should inflate the noise term (reference) and be either user specified or can be adaptively chosen via a simple scheme.

 (Note choosing the step automatically will involve evaluating the forward model, the user should be aware of this if the forward model is expensive)

PR #60, opened 2020-06-26
Add timestep for EKI, and timestep finder function.

ilopezgp:
Added timestep for EKI and a function to find the EKI step based on the reduction of the ensemble covariance determinant after one stage of EKI.

An additional warning message has also been implemented to the EKI update in the case of rapid collapse of the ensemble.

This PR fixes issue #56

### DIFF
--- a/src/EKI.jl
+++ b/src/EKI.jl
@@ -95,7 +95,7 @@ function compute_error(eki)
 end
 
 
-function update_ensemble!(eki::EKIObj{FT}, g; Δt_new = nothing) where {FT}
+function update_ensemble!(eki::EKIObj{FT}, g; cov_threshold::FT=0.01, Δt_new = nothing) where {FT}
     # u: N_ens x N_params
     u = eki.u[end]
     cov_init = cov(eki.u[end], dims=1)
@@ -149,10 +149,10 @@ function update_ensemble!(eki::EKIObj{FT}, g; Δt_new = nothing) where {FT}
 
     # Check convergence
     cov_new = cov(eki.u[end], dims=1)
-    cov_threshold = 0.01
-    if det(cov_new) < cov_threshold*det(cov_init)
-        @warn "Ensemble covariance after the last EKI stage decreased significantly. 
-            Consider adjusting the EKI time step."
+    cov_ratio = det(cov_new)/det(cov_init)
+    if cov_ratio < cov_threshold
+        @warn string("New ensemble covariance determinant is less than ",cov_threshold," times its former value.
+            Consider reducing the EKI time step.")
     end
 end
 
--- a/src/EKI.jl
+++ b/src/EKI.jl
@@ -36,6 +36,8 @@ struct EKIObj{FT<:AbstractFloat, IT<:Int}
      g::Vector{Array{FT, 2}}
      "vector of errors"
      err::Vector{FT}
+     "vector of timesteps used in each EKI iteration"
+     Δt::Vector{FT}
 end
 
 # outer constructors
@@ -54,8 +56,10 @@ function EKIObj(parameters::Array{FT, 2},
     g = Vector{FT}[]
     # error store
     err = []
+    # timestep store
+    Δt = []
 
-    EKIObj{FT,IT}(u, parameter_names, t_mean, t_cov, N_ens, g, err)
+    EKIObj{FT,IT}(u, parameter_names, t_mean, t_cov, N_ens, g, err, Δt)
 end
 
 
@@ -91,7 +95,7 @@ function compute_error(eki)
 end
 
 
-function update_ensemble!(eki::EKIObj{FT}, g; Δt=FT(1.0)) where {FT}
+function update_ensemble!(eki::EKIObj{FT}, g; Δt_new = nothing) where {FT}
     # u: N_ens x N_params
     u = eki.u[end]
     cov_init = cov(eki.u[end], dims=1)
@@ -102,6 +106,14 @@ function update_ensemble!(eki::EKIObj{FT}, g; Δt=FT(1.0)) where {FT}
 
     cov_ug = fill(FT(0), size(u)[2], size(g)[2])
     cov_gg = fill(FT(0), size(g)[2], size(g)[2])
+
+    if !isnothing(Δt_new)
+        push!(eki.Δt, Δt_new)
+    elseif isnothing(Δt_new) && isempty(eki.Δt)
+        push!(eki.Δt, FT(1))
+    else
+        push!(eki.Δt, eki.Δt[end])
+    end
 
     # update means/covs with new param/observation pairs u, g
     for j = 1:eki.N_ens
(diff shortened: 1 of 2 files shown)


src/EKI.jl
```diff
@@ -91,9 +91,10 @@ function compute_error(eki)
 end
 
 
-function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
+function update_ensemble!(eki::EKIObj{FT}, g; Δt=1.0) where {FT}
```
charleskawczynski, line 94:
```suggestion
function update_ensemble!(eki::EKIObj{FT}, g; Δt=FT(1.0)) where {FT}
```

src/EKI.jl
```diff
@@ -134,6 +135,39 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
...
+
+
+"""
+   find_eki_step(eki::EKIObj{FT}, g::Array{FT, 2}; cov_threshold::FT=0.01) where {FT}
+Find largest step for the EKI solver that leads to a reduction of the determinant of the sample  
+covariance matrix no greater than cov_threshold.
+"""
+function find_eki_step(eki::EKIObj{FT}, g::Array{FT, 2}; cov_threshold::FT=0.01) where {FT}
+    accept_step = false
+    Δt = 1.0
```
charleskawczynski, line 155:
```suggestion
    Δt = FT(1.0)
```

src/EKI.jl
```diff
@@ -134,6 +135,39 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
...
+    # u: N_ens x N_params
+    cov_init = cov(eki.u[end], dims=1)
+    while accept_step == false
+        eki_copy = deepcopy(eki)
+        update_ensemble!(eki_copy, g, Δt=Δt)
+        cov_new = cov(eki_copy.u[end], dims=1)
+        if det(cov_new) > cov_threshold*det(cov_init)
+            accept_step = true
+        else
+            Δt = Δt/2.0
```
charleskawczynski, line 165:
```suggestion
            Δt = Δt/2
```

src/EKI.jl
```diff
@@ -134,6 +135,39 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
 
     compute_error(eki)
 
+    # Check convergence
+    cov_new = cov(eki.u[end], dims=1)
+    cov_threshold = 0.01
+    if det(cov_new) < cov_threshold*det(cov_init)
+        println("Warning: Ensemble covariance after the last EKI stage decreased significantly.
```
charleskawczynski, line 142:
```suggestion
        @warn "Ensemble covariance after the last EKI stage decreased significantly."
```

src/EKI.jl
```diff
@@ -134,6 +135,39 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
 
     compute_error(eki)
 
+    # Check convergence
+    cov_new = cov(eki.u[end], dims=1)
+    cov_threshold = 0.01
+    if det(cov_new) < cov_threshold*det(cov_init)
+        println("Warning: Ensemble covariance after the last EKI stage decreased significantly. 
+            Consider adjusting the EKI time step.")
```
charleskawczynski, line 143:
```suggestion
        @warn "Ensemble covariance after the last EKI stage decreased significantly. 
            Consider adjusting the EKI time step."
```

odunbar:
It would be good to store the timestep - so we can look at it from `EKIObj` after the simulation - especially as it is chosen adaptively. 

It is also possible in the framework you have written that a user could choose a new timestep between every update call. So perhaps the best way is to store the timestep for each iteration.

odunbar:
We could also have a short example along the lines already written just for the EKI stage?

1. Create a near linear problem
2. Perform an EKI step leading to collapse. (to display the warning)
3. With a new EKI object perform the stepfinder to choose a timestep size
4. Perform the new EKI Step.

This should be be quite easy from the the example you already have created.

charleskawczynski:
> @charleskawczynski Where do you think the best place for small modular tests would be? I feel these should be distinct from the 'Cloudy' example etc.

[Here](https://github.com/CliMA/CalibrateEmulateSample.jl/blob/master/test/EKI) and include it in `runtests.jl` or in [runtests.jl](https://github.com/CliMA/CalibrateEmulateSample.jl/blob/master/test/EKI/runtests.jl) directly.

ilopezgp:
> It would be good to store the timestep - so we can look at it from `EKIObj` after the simulation - especially as it is chosen adaptively.
> 
> It is also possible in the framework you have written that a user could choose a new timestep between every update call. So perhaps the best way is to store the timestep for each iteration.

Sounds good. I will store a vector containing the timestep for each stage, similar to how the other variables are stored.

ilopezgp:
> We could also have a short example along the lines already written just for the EKI stage?
> I'm thinking something along the lines of:
> 
> 1. Create a nearly linear problem
> 2. Perform an EKI step leading to collapse. (to display the warning)
> 3. With a new EKI object perform the stepfinder to choose a timestep size. Print it out by calling the stored value `EKIObj.DeltaT`
> 4. Perform the new EKI Step(s). Succesfully. and print out the cov and new `EKIObj,DeltaT`
> 
> This should be be quite easy from the the example you already have created.
> 
> @charleskawczynski Where do you think the best place for small modular tests would be? I feel these should be distinct from the 'Cloudy' example etc.

Would this be an example or a test then? Should it be included in this PR?

charleskawczynski:
> Would this be an example or a test then? Should it be included in this PR?

A test, and yes, it would be good to include in this PR.

src/EKI.jl
```diff
@@ -54,8 +56,10 @@ function EKIObj(parameters::Array{FT, 2},
     g = Vector{FT}[]
     # error store
     err = []
+    # timestep store
+    Δt = []
```
charleskawczynski, line 60:
```suggestion
    Δt = FT[]
```

src/EKI.jl
```diff
@@ -134,6 +147,43 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
 
     compute_error(eki)
 
+    # Check convergence
+    cov_new = cov(eki.u[end], dims=1)
+    cov_threshold = 0.01
```
bielim, line 152: Should we make `cov_threshold` an input argument to `update_ensemble!` (with a default value)? 
At the moment I can see two different ways users will run an EKI:

1. Directly apply `update_ensemble!` without doing a step size search first. In this case the user will get a warning if the ensemble collapses faster than the rate implied by the hard-coded `cov_threshold`.
2. Run `find_eki_step` to find a good step size before the call to `update_ensemble!`. But with the current implementation it's possible that the user is only concerned about collapse speeds greater than what triggers the warning in `update_ensemble`, so he would get the warning despite having chosen a step size that's consistent with the maximum collapse speed he is willing to accept. Conversely, a user might want the ensemble to collapse more slowly and would like to see a warning already at a smaller decrease of the ensemble covariance than what's currently implemented.

src/EKI.jl
```diff
@@ -134,6 +147,43 @@ function update_ensemble!(eki::EKIObj{FT}, g) where {FT}
...
+    if det(cov_new) < cov_threshold*det(cov_init)
+        @warn "Ensemble covariance after the last EKI stage decreased significantly.
```
bielim, line 154: No matter whether `cov_threshold` ends up being hard-coded or not, I would replace "significantly" with the actual quantitative information about how much the covariance has decreased in that iteration.

src/EKI.jl
ilopezgp: You are right, I can see how it would be annoying to get the warning even when the user wants fast convergence (e.g. just for calibrating). I will modify this by adding `con_threshold` as a kwarg set to 0.01 by default.

src/EKI.jl
ilopezgp: I agree, specially after your previous comment and allowing a user based `con_threshold`.

PR #62, opened 2020-06-27
Extend tests for GPEmulator.jl and Observations.jl

bielim:
This PR extends the coverage of the tests for the modules `GPEmulator.jl` and `Observations.jl`.

### DIFF
--- a/src/GPEmulator.jl
+++ b/src/GPEmulator.jl
@@ -117,12 +117,12 @@ function GPObj(inputs, data, package::GPJL; GPkernel::Union{K, KPy, Nothing}=not
         GPkernel_i = deepcopy(GPkernel)
         # inputs: N_samples x N_parameters
         # data: N_samples x N_data
-        logstd_obs_noise = log(sqrt(0.5)) # log standard dev of obs noise
+        logstd_obs_noise = log(sqrt(0.01)) # log standard dev of obs noise
         # Zero mean function
         kmean = MeanZero()
         m = GPE(inputs', dropdims(data[:, i]', dims=1), kmean, GPkernel_i, 
                 logstd_obs_noise)
-        optimize!(m, noise=false)
+        optimize!(m, noise=true)
         push!(models, m)
     end
     return GPObj{FT, typeof(package)}(inputs, data, input_mean, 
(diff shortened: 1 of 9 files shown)


src/GPEmulator.jl
```diff
@@ -117,12 +117,12 @@ function GPObj(inputs, data, package::GPJL; GPkernel::Union{K, KPy, Nothing}=not
         GPkernel_i = deepcopy(GPkernel)
         # inputs: N_samples x N_parameters
         # data: N_samples x N_data
-        logstd_obs_noise = log(sqrt(0.5)) # log standard dev of obs noise
+        logstd_obs_noise = log(sqrt(0.01)) # log standard dev of obs noise
```
ilopezgp, line 120: Should we make this dependent on the ensemble covariance? I also think that Ollie wanted to base this on the SVD stuff.

src/Observations.jl
```diff
@@ -70,16 +69,23 @@ function Obs(samples::Vector{Vector{FT}},
              cov::Array{FT, 2},
              data_names::Vector{String}) where {FT<:AbstractFloat}
     N_samples = length(samples)
+    N_data = length(samples[1]) # N data points per observation sample
+    # convert to N_samples x N_data to determine sample covariance
```
charleskawczynski, line 73:
```suggestion
```

src/GPEmulator.jl
bielim: Yes, we should! This will come in another (probably the next) PR though - the purpose of this one is merely to extend the test coverage. The change of the hard-coded regularization noise from `log(sqrt(0.5))` to `log(sqrt(0.01))` is just there because I experimented with it after our discussion and thought I could as well leave it at a smaller value until we have implemented the SVD (since the idea is that the regularization noise should be smaller than the learned white noise, a criterion that is more likely to be met the smaller the regularization noise is - speaking of which, one of the next PRs should also address the issue of naming all these noises (white noise, observational noise, regularization noise etc.) in a consistent way that minimizes confusion as much as possible).

ilopezgp: Sounds good. And yes, I think with all the covariances involved, being precise in the notation would be helpful.

src/GPEmulator.jl
```diff
@@ -117,12 +117,12 @@ function GPObj(inputs, data, package::GPJL; GPkernel::Union{K, KPy, Nothing}=not
...
         # Zero mean function
         kmean = MeanZero()
         m = GPE(inputs', dropdims(data[:, i]', dims=1), kmean, GPkernel_i, 
                 logstd_obs_noise)
-        optimize!(m, noise=false)
+        optimize!(m, noise=true)
```
ilopezgp, line 125: FYI this is the default of `optimize!`. I will try to read the documentation of GaussianProcesses.jl to see what would be the best setting for our purposes.

PR #63, opened 2020-06-29
Add SVD functionality

### LINKED ISSUE #57: SVD decomposition
In the papers to be released on the GCM use of CES we use an SVD decomposition to transform the space in which we learn with GP. I suggest we implement this here.

Primarily this allows the learning of noise to be exact and reguralization to be precise, as we transform into a space where the observational noise is not only diagonal but we can normalize it to be exactly the identity. Then the built in `alpha` or `observational_noise` parameters in the GP are more naturally chosen.

This will involve firstly an SVD implementation within the `GPEmulator` class, and secondly a modified `MCMC` class to work in the transformed variables.

### LINKED ISSUE #58: Noise learning bug in GP
In some recent testing there appeared to be some bugs in the white noise kernel in `GPemulator`.

https://github.com/CliMA/CalibrateEmulateSample.jl/blob/5fb71c2109b7ff818ceb8339f3351f654234aecc/src/GPEmulator.jl#L111

Firstly, if users know the noise structure completely, they can provide the covariance matrix as an optional argument to `GPEmulator`.

Secondly, if the noise structure is unknown, then the white noise kernel should be used to learn the noise.

@ilopezgp has provided a simple working example where learning the observational noise fails even in a simple diagonal case.

odunbar:
Resolves #57 . We add the ability for the Gaussian process to learn non - diagonal covariance matrix for the observational noise by applying and SVD.

Resolve #58 We add functionality to toggle whether we wish to learn the observational noise. and add (mathematically correct) default values when we are not.

We also have 2 examples:

- [x] Gaussian process plot. We train a GP on some training points in 2D space, and plot the mean and variance (compared with the underlying model and observational noise) - in the untransformed space

- [x] Noise learning test. We train a GP with known noise, with `learn_noise = false` and with `learn_noise = true` we compare the learned `WhiteKernel`  parameters to the true unlearned parameters.

Coauthored with @bielim


src/GPEmulator.jl
```diff
@@ -64,28 +64,49 @@ struct GPObj{FT<:AbstractFloat, GPM}
     models::Vector
     "whether to fit GP models on normalized inputs ((inputs - input_mean) * sqrt_inv_input_cov)"
     normalized::Bool
+    "the singular value decomposition matrix"
+    decomposition
+    "the normalization in the svd transformed space"
+    sqrt_singular_values
```
ilopezgp, line 70: Can we add the types of these attributes?

src/GPEmulator.jl
```diff
@@ -95,42 +116,81 @@ function GPObj(inputs, data, package::GPJL; GPkernel::Union{K, KPy, Nothing}=not
         sqrt_inv_input_cov = convert(Array{FT}, sqrt(inv(cov(inputs, dims=1))))
         inputs = (inputs .- input_mean) * sqrt_inv_input_cov
     end
-        
+
+    # Transform the outputs
+    sqrt_singular_values_inv = Diagonal(1.0 ./ sqrt.(SVD.S))
```
ilopezgp, line 121: Should `sqrt.(SVD.S)` be `sqrt.(decomposition.S)`?

src/GPEmulator.jl
odunbar: Done.

src/GPEmulator.jl
odunbar: Done

bielim:
I just added an example of how to fit and plot a GP (in examples/GPEmulator). The code should run through and generate plots, but there seems to be something wrong with the actual GP regression that gets fit to the data - e.g., compare the truth to the mean predicted by the GP; this is clearly not a good fit. The same conclusion can be reached by probing the GP predictions at a couple of individual test points.
Any suggestions on what might be going wrong are welcome!

bielim:
Also: I tried to change the `predict` function such that it can make predictions on more than one point at a time. While the ability to predict at a single new test point is all that's needed for use of the GP emulator within the MCMC, it would be nice to support predictions at multiple points at once (i.e., the `predict` function should work on inputs of size n_test_points x n_parameters). This requires some extra manipulations when the `transform_to_real` flag is set to `true`: the result of the back transformation of the variance (which involves a diagonal matrix with the predicted (transformed) variances as its entries) is a square matrix, but the expected return value from the `predict` function are the means and variances of the test points, so the variances have to be re-extracted from the square matrix (or matrices, if we're dealing with multiple test points) resulting from the transformation(s). 
I'm not sure at all if I got this right, please check :-) (however, the plotted results don't look better with the `transform_to_real` set to `false`, so even if there's something wrong with the back transformation, it can't be the only source of error).

examples/GPEmulator/plot_GP.jl
```diff
@@ -0,0 +1,94 @@
...
+    plot(p1, p2, layout=(1, 2), legend=false)
+    savefig("/home/melanie/Desktop/GP_test_y"*string(y_i)*".png")
+end
+
+# Make plots of the true components of G(x1, x2)
+g1_true = sin.(inputs[:, 1]) .+ cos.(inputs[:, 2])
+g1_grid = reshape(g1_true, n_pts, n_pts)
+p3 = plot(x1, x2, g1_grid, st=:surface, camera=(-30, 30), c=:cividis, 
+          xlabel="x1", ylabel="x2", zlabel="sin(x1) + 0.5*x2")
+savefig("/home/melanie/Desktop/GP_test_true_g1.png")
```
charleskawczynski, line 88:
```suggestion
savefig("GP_test_true_g1.png")
```

examples/GPEmulator/plot_GP.jl
```diff
@@ -0,0 +1,94 @@
...
+
+g2_true = sin.(inputs[:, 1]) .- cos.(inputs[:, 2])
+g2_grid = reshape(g2_true, n_pts, n_pts)
+p4 = plot(x1, x2, g2_grid, st=:surface, camera=(-30, 30), c=:cividis, 
+          xlabel="x1", ylabel="x2", zlabel="sin(x1) + 0.5*x2")
+savefig("/home/melanie/Desktop/GP_test_true_g2.png")
```
charleskawczynski, line 94:
```suggestion
savefig("GP_test_true_g2.png")
```

examples/GPEmulator/plot_GP.jl
```diff
@@ -0,0 +1,94 @@
...
+    mean_grid = reshape(gp_mean[y_i, :], n_pts, n_pts)
+    p1 = plot(x1, x2, mean_grid, st=:surface, camera=(-30, 30), c=:cividis, 
+              xlabel="x1", ylabel="x2", zlabel="mean of y"*string(y_i))
+
+    var_grid = reshape(gp_var[y_i, :], n_pts, n_pts)
+    p2 = plot(x1, x2, var_grid, st=:surface, camera=(-30, 30), c=:cividis,
+              xlabel="x1", ylabel="x2", zlabel="var of y"*string(y_i))
+
+    plot(p1, p2, layout=(1, 2), legend=false)
+    savefig("/home/melanie/Desktop/GP_test_y"*string(y_i)*".png")
```
charleskawczynski, line 80:
```suggestion
    savefig("GP_test_y"*string(y_i)*".png")
```

ilopezgp:
> Also: I tried to change the `predict` function such that it can make predictions on more than one point at a time. While the ability to predict at a single new test point is all that's needed for use of the GP emulator within the MCMC, it would be nice to support predictions at multiple points at once (i.e., the `predict` function should work on inputs of size n_test_points x n_parameters). This requires some extra manipulations when the `transform_to_real` flag is set to `true`: the result of the back transformation of the variance (which involves a diagonal matrix with the predicted (transformed) variances as its entries) is a square matrix, but the expected return value from the `predict` function are the means and variances of the test points, so the variances have to be re-extracted from the square matrix (or matrices, if we're dealing with multiple test points) resulting from the transformation(s).
> I'm not sure at all if I got this right, please check :-) (however, the plotted results don't look better with the `transform_to_real` set to `false`, so even if there's something wrong with the back transformation, it can't be the only source of error).

Have you added any of these changes to your commit? The only change I see is the addition of the new example.

examples/GPEmulator/plot_GP.jl
```diff
@@ -0,0 +1,94 @@
...
+g1 = sin.(X[:, 1]) .+ cos.(X[:, 2])
+g2 = sin.(X[:, 1]) .- cos.(X[:, 2])
+Y = [g1 g2] .+ rand(MvNormal(μ, Σ), n)' 
+truth_cov = cov(Y, dims=1)
+
+# Fit 2D Gaussian Process regression model
+# (To be precise, we fit two models, one that predicts y1 from x1 and x2, 
+# and one that predicts y2 from x1 and x2)
+gpobj = GPObj(X, Y, truth_cov, gppackage, GPkernel=nothing, 
+              normalized=true, noise_learn=true, prediction_type=pred_type)
```
ilopezgp, line 55: Since this is a tutorial for fitting a Gaussian process, it would be nice to either discuss what the default GPKernel is or to define and use a simple one here.

examples/GPEmulator/plot_GP.jl
bielim: Oops, forgot to change that to a "neutral" path for the PR :-) Thanks!

examples/GPEmulator/plot_GP.jl
bielim: Yes, agreed. I'll add either some explanation or demonstrate how to construct your own kernel once the example works.

bielim:
> > Also: I tried to change the `predict` function such that it can make predictions on more than one point at a time. While the ability to predict at a single new test point is all that's needed for use of the GP emulator within the MCMC, it would be nice to support predictions at multiple points at once (i.e., the `predict` function should work on inputs of size n_test_points x n_parameters). This requires some extra manipulations when the `transform_to_real` flag is set to `true`: the result of the back transformation of the variance (which involves a diagonal matrix with the predicted (transformed) variances as its entries) is a square matrix, but the expected return value from the `predict` function are the means and variances of the test points, so the variances have to be re-extracted from the square matrix (or matrices, if we're dealing with multiple test points) resulting from the transformation(s).
> > I'm not sure at all if I got this right, please check :-) (however, the plotted results don't look better with the `transform_to_real` set to `false`, so even if there's something wrong with the back transformation, it can't be the only source of error).
> 
> Have you added any of these changes to your commit? The only change I see is the addition of the new example.

Yes, they are there now, they got added with commit 1661de0

ilopezgp:
> > > Also: I tried to change the `predict` function such that it can make predictions on more than one point at a time. While the ability to predict at a single new test point is all that's needed for use of the GP emulator within the MCMC, it would be nice to support predictions at multiple points at once (i.e., the `predict` function should work on inputs of size n_test_points x n_parameters). This requires some extra manipulations when the `transform_to_real` flag is set to `true`: the result of the back transformation of the variance (which involves a diagonal matrix with the predicted (transformed) variances as its entries) is a square matrix, but the expected return value from the `predict` function are the means and variances of the test points, so the variances have to be re-extracted from the square matrix (or matrices, if we're dealing with multiple test points) resulting from the transformation(s).
> > > I'm not sure at all if I got this right, please check :-) (however, the plotted results don't look better with the `transform_to_real` set to `false`, so even if there's something wrong with the back transformation, it can't be the only source of error).
> > 
> > 
> > Have you added any of these changes to your commit? The only change I see is the addition of the new example.
> 
> Yes, they are there now, they got added with commit [1661de0](https://github.com/CliMA/CalibrateEmulateSample.jl/commit/1661de0d8e7f61e8eb54db4b1705e54014f05f5d)

Can we test this with a working example before merging? It should output the same result as the former formulation called for all the test points in the array.

odunbar:
>Can we test this with a working example before merging? It should output the same result as the former formulation called for all the test points in the array.

Yeah, don't worry this PR is listed with WIP -> "work in progress" flag. There is no intention of merging before this is properly tested. I am looking into Melanie's prediction issues now.

bielim:
> > Can we test this with a working example before merging? It should output the same result as the former formulation called for all the test points in the array.
> 
> Yeah, don't worry this PR is listed with WIP -> "work in progress" flag. There is no intention of merging before this is properly tested. I am looking into Melanie's prediction issues now.

Exactly! Sorry, @ilopezgp, we should have been clear about the meaning of WIP. Of course things will only be merged after testing - merging is a bad idea when your comment to what you committed ends with the sentence "Any suggestions on what might be going wrong are welcome!" :-)

bielim:
Okay, as of commit #62838e3 the `plot_GP.jl` example produces the correct output. So far it only works for `gppackage=GPJL()`. Attempts to use `ScikitLearn` (`gppackage=SKLJL()`) result in a seg fault whose cause I do not understand. The seg fault doesn't seem to be related to wrong input dimensions or the like, it occurs as soon as `GPEmulator` makes any attempt at calling one of the functions imported via `sk_import`.
My next steps:
- [ ] Change the output of the `predict` function to return the full covariances rather than just the variances. 
- [ ] Use multiple dispatch to handle different output and input dimensions: E.g., the code currently doesn't work for 1D output, for which no noise covariance is available and therefore no SVD decomposition can be done. Or, currently only 2D-inputs are accepted, so the user has to make sure that e.g. row vectors (rank-1 arrays) are reshaped to 1xN matrices, and that scalars are reshaped to 1x1 matrices - this is not very user-friendly.
- [ ] Add an example that demonstrates the `noise_learn` functionality.

bielim:
As of commit #5c7e89c, the code passes all tests (some of which I have extended). There have been a number of changes in terms of input-output handling, some renaming, clean-up, etc.
**Note**: Originally the plan was to use multiple dispatch to handle different output and input dimensions, e.g., vectors (rank-1 arrays) and scalars instead of 2D arrays. I went back and forth on this but eventually decided to only allow 2D arrays as input and output to `GPObj`. This means that vectors have to be reshaped to 2D arrays of size (1, N_inputs) or (1, N_outputs). While this may seem tedious, it allows to clearly distinguish the case where a user passes N 1-dim (i.e., scalar) samples from the case where a user passes 1 N-dim sample. The former case should be passed as a N x 1 array (a column vector), whereas the latter case should be passed as a 1 x N array (a row vector).
I'd appreciate if people could have a look at the changes, and it would also be good to have a discussion about the current status of this PR and the remaining steps before it can be merged.

Project.toml
```diff
@@ -13,9 +13,12 @@ GaussianProcesses = "891a1506-143c-57d2-908e-e1f8e92e6de9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
charleskawczynski, line 16: Likely, CES.jl shouldn't directly depend on Plots (#65). Rather, we can make a test environment (https://julialang.github.io/Pkg.jl/dev/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1).

odunbar: Yup this is what we are thinking - We are planning on merging this PR first, then doing a 'CES' cleanup to deal with the `project.toml` 's and alike to put dependencies only where required

bielim: @charleskawczynski, that's right, but we will address #65 with a separate PR following this one. In that "clean-up PR", we will create separate `Project.toml` and `Manifest.toml` files for each example, with the idea that this will get rid of some annoying dependencies that are only needed for running the examples, but not for the core CES functionality. There are also quite a few modules in `src/` that are not functional and are not being developed any further -- removing them will be part of the cleaning as well.

bielim: My comment came too late :-)

odunbar:
bors try

Issue #65, opened 2020-07-07
Insanely long precompile times.


odunbar:
My precompile time is around 10 minutes or more for the package on my laptop (which is still a reasonably good machine). The CES julia code i've used before has a 30 second precompile time...

 I can limit this with `Revise.jl` but it's quite unusable otherwise.

jakebolewski:
When I was updating the CI this took about 17 minutes for me.

I think the only short term practical solution here is to break apart the package to cut down on the number of third party dependencies / and or have a custom sys-image for the dependencies similar to: https://github.com/CliMA/ClimateMachine.jl/tree/master/.dev/systemimage

odunbar:
Yes I think we can strip it out of those, But it's surprising - I have a similar code (which this was based on) and that seems to take on the order of 30seconds to precompile. I'll maybe have a look for differences.

jakebolewski:
Given the way (current) pre-compilation works, you can get into scenarios where there is excessive method invalidation given the mix of packages imported and the methods they extend (one package invalidates a pre-compiled method in another which has cascading consequences).

odunbar:
Is there a way of automatically flushing out deprecated packages from the `project.tml` / `manifest.toml` ?

jakebolewski:
unfortunately no

odunbar:
haha well I guess we will find a way then ;)

odunbar:
Actually, is there a way of isolating certain packages with certain examples then? 

Basically I think the structure is that we have some core packages required by the `src` files (e.g. `LinearAlgebra` ).

But then we also have some `example` files, and for each example we may have 3-4 different packages loaded that are only to run the example (e,g `DifferentialEquations`). These are only to be called when the example is run, but the precompilation will compile everything at once. I'm wondering if we can create these examples as `modules`  which are only compiled when the example is run?

glwagner:
@odunbar you may be able to use test dependencies for the examples.

We use something similar in `Oceananigans.jl`. The examples use `Plots`. Thus to run examples during tests, we permit the tests to depend on `Plots`, even though `Oceananigans` does not depend on `Plots`:

https://github.com/CliMA/Oceananigans.jl/blob/master/Project.toml

Some more info here:

https://discourse.julialang.org/t/how-to-add-a-package-as-a-test-dependency-from-the-repl/14313

odunbar:
Thank Greg! I've had a look. 

[...]

PR #67, opened 2020-07-22
Clean up CalibrateEmulateSample

bielim:
This PR is a major clean-up (goal: code hygiene and reduction of the insanely long precompile times, i.e., addressing #65) -- the following changes were made:

(1) Deleted the modules `EKS.jl`, `GPR.jl`, `Histograms.jl`, `problems.jl`, and `spaces.jl`. These modules are not functional and not being developed anymore.

(2) Deleted `test/EKS`, `test/Histograms`, `test/GPR`,  `test/L96m`, as well as `examples/GPR` and `examples/L96`, because they are based on non-functional modules or (in the case of `L96`) because they don't seem to demonstrate any CES functionality.

(3) Deleted `test/Cloudy`, moved and renamed its content (`test/Cloudy/runtests.jl`) to `examples/Cloudy/Cloudy_example.jl`. The Cloudy "test" never was a proper test but rather an example, so it fits much better in the `examples` folder. The Cloudy notebook (`examples/Cloudy/calibrate_emulate_sample_Cloudy_demo.ipynb`) was deleted as it's basically just a more "eye-friendly" version of `Cloudy_example.jl`.

(4) Moved `GModel.jl` to `examples/Cloudy`.  This module shows what a forward model G that maps the parameters u to G(u) could look like, but it does so using Cloudy as an example, and for that reason I never liked the fact that this module was there with all the general-purpose code in the first place. As a result of (3) and (4), there is no more Cloudy dependency in the tests or core CES code.

(5) Generated separate `Project.toml` and `Manifest.toml` files for each example. With this setup, running the core CES code does not require to precompile packages that are only used in the examples (e.g., `DifferentialEquations.jl`) anymore.

(6) Removed unused packages from the main `Project.toml` file

Finally, I also changed the way the `ScikitLearn.jl` imports are handled in `GPEmulator.jl`. At least on my machine, this solves the problems with the seg faults that I mentioned in the discussion to PR #63.


bielim:
@dburov190: I believe you are the creator of `L96` -- I can see that it is an implementation of the Lorenz 96 model, but I'm not sure if it demonstrates any CES functionality. If it does, please let me know, and I'll re-include it. If not, I'd suggest to find another place/repo for it, in the interest of keeping the CES repo as clean as possible.

bielim:
@charleskawczynski: One issue I see with the current structure (separate `Project.toml` files for each example while keeping only the packages needed for core CES functionality in the main `Project.toml`) is that it is very easy to accidentally add some of the example dependencies back in to the main `Project.toml`: E.g., when I am in`CalibrateEmulateSample.jl/` and do `include("examples/Cloudy/Cloudy_example.jl")` (rather than running the example from within `examples/Cloudy`), it will automatically start to update the main `Project.toml` with the packages needed to run `Cloudy_example.jl`. Is there a way to keep this from happening?

bielim:
bors try

jakebolewski:
35 -> 13 minutes, nice!

charleskawczynski:
@bielim looks like you figured out how to avoid this auto-project update problem. Please reach out if this is not the case!

PR #41, opened 2020-02-18
Added Julia version for EKS

charleskawczynski:
Copied from #40

PR #70, opened 2020-07-24
WIP: Add Ensemble Kalman Sampler

bielim:
This PR adds a Julia implementation of an Ensemble Kalman Sampler. It supersedes PR #41.


bielim:
@agarbuno: This PR is based on your EKS implementation. However, the EKS update needs some revision: First, it errored due to dimensional mismatches, which led me to look at the implicit equation implemented in lines 250-256 in `src/EKP.jl`. That equation seems to differ from equation (2.8a) in the [Calibrate-Emulate-Sample paper](https://arxiv.org/pdf/2001.03689.pdf), and so I didn't know how to correctly adjust it. Would be great if you could have a look at it!
The current implementation runs through (thanks to replacing the `+` in the beginning of line 253 with a `.+`, but it doesn't produce a correct EKS iteration (i.e., I ran the EKS test case implemented in `test/EKP/runtests.jl` and plotted the evolution of the particles in each iteration -- the particles are diverging away from the initial ensemble rather than being distributed in regions of high posterior probability).

examples/Cloudy/Cloudy_example.jl
```diff
@@ -118,8 +118,8 @@ exp_transform(a::AbstractArray) = exp.(a)
 N_ens = 50 # number of ensemble members
 N_iter = 5 # number of EKI iterations
 # initial parameters: N_ens x N_params
-initial_params = EKI.construct_initial_ensemble(N_ens, priors; rng_seed=6)
-ekiobj = EKI.EKIObj(initial_params, param_names, truth.mean, truth.obs_noise_cov, Δt=1.0)
+initial_params = EKP.construct_initial_ensemble(N_ens, priors; rng_seed=6)
+ekiobj = EKP.EKObj(initial_params, param_names, truth.mean, truth.obs_noise_cov, Δt=1.0)
```
ilopezgp, line 122: Is this missing the new Process argument? I believe there is no default constructor for this argument

bielim: Yes, thanks! Fixed it. The original implementation of the new `EKObj` had Inversion as the default process, and the omission of the Process argument was a leftover from that. In general I am still not sure whether it would be better to have separate constructors for Inversion EKObjs and Sampler EKObjs, as the optional time step argument is unused (and hence potentially confusing) for EK Samplers...

PR #72, opened 2020-07-30
add a new calibration method UKI (unscented Kalman inversion) and an example Lorenz63

Zhengyu-Huang:
Add new file src/UKI.jl
Modify CalibrateEmulateSample.jl to include UKI.jl
Add a test example for UKI Lorenz63 in the examples folder

### DIFF
--- a/examples/Lorenz63/Lorenz63.jl
+++ b/examples/Lorenz63/Lorenz63.jl
@@ -1,0 +1,188 @@
+using JLD2
+using Statistics
+using LinearAlgebra
+using PyPlot
+using CalibrateEmulateSample.UKI
+
+
+function f(x::Array{Float64,1}, σ::Float64, r::Float64, β::Float64)
+    return [σ*(x[2]-x[1]); x[1]*(r-x[3])-x[2]; x[1]*x[2]-β*x[3]] 
+end
+
+
+function compute_foward(μ::Array{Float64,1}, Δt::Float64, nt::Int64)
+    @assert(length(μ) == 6)
+    nx = 3
+    x0 = μ[1:nx]
+    σ,r, β = μ[nx+1:6]
+    
+    xs = zeros(nx, nt+1)
+    xs[:,1] = x0
+    for i=2:nt+1
+        xs[:,i] = xs[:,i-1] + Δt*f(xs[:,i-1], σ, r, β)
+    end
+    
+    return xs
+end
+
+
+
+function compute_obs(xs::Array{Float64,2})
+    # N_x × N_t
+    x1, x2, x3 = xs[1,:]', xs[2,:]', xs[3,:]'
+    obs = [x1; x2; x3; x1.^2; x2.^2; x3.^2]
+end
+
+# fix σ, learn r and β
+# f = [σ*(x[2]-x[1]); x[1]*(r-x[3])-x[2]; x[1]*x[2]-β*x[3]]
+# σ, r, β = 10.0, 28.0, 8.0/3.0
+function run_Lorenz_ensemble(params_i, Tobs, Tspinup, Δt)
+    T = Tobs + Tspinup
+    nt = Int64(T/Δt)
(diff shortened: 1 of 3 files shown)


ilopezgp:
@Zhengyu-Huang @bielim this PR and #70 seem to be going in different direction (merging all calibrating methods in one src file versus many). It might be a good idea to meet and discuss how to move forward with the code.

src/UKI.jl
```diff
@@ -0,0 +1,242 @@
...
+Struct that is used in Unscented Kalman Inversion (UKI)
+
+#Fields
+$(DocStringExtensions.FIELDS)
+"""
+struct UKIObj{FT<:AbstractFloat, IT<:Int}
+    "vector of parameter names (never used)"
+     unames::Vector{String}
+    "a vector of arrays of size N_ensemble x N_parameters containing the mean of the parameters (in each UKI iteration a new array of mean is added)"
+     θ_bar::Vector{Array{FT}}
```
ilopezgp, line 20: The descriptor says that the size of each vector element is N_ensemble x N_parameters, but `θ_bar::Vector{Array{FT}}` is a vector of 1D arrays.

ilopezgp: It might also be useful for integration with EKI and EKS if all these objects have the same attributes. Here you are storing the mean and covariance, whereas in `EKI.jl` we store all the parameters. Storing the parameters is useful when considering the emulation stage.

src/UKI.jl
```diff
@@ -0,0 +1,242 @@
...
+     "parameter size g = G(θ)"
+     N_θ::IT
+     "observation size g = G(θ)"
+     N_g::IT
+     "weights in UKI"
+     sample_weights::Array{FT}
+     μ_weights::Array{FT} 
+     cov_weights::Array{FT} 
+     "regularization parameter"
+     α_reg::Float64
```
ilopezgp, line 40: α_reg::FT

src/UKI.jl
```diff
@@ -0,0 +1,242 @@
...
+θ0_bar::Array{FT} : prior mean
+θθ0_cov::Array{FT, 2} : prior covariance
+g_t::Vector{FT} : observation 
+obs_cov::Array{FT, 2} : observation covariance
+α_reg::Float64 : regularization parameter toward θ0 (0 < α_reg <= 1), default should be 1, without regulariazion
+"""
+function UKIObj(parameter_names::Vector{String},
+                θ0_bar::Array{FT}, 
+                θθ0_cov::Array{FT, 2},
+                g_t::Vector{FT}, # observation
```
ilopezgp, line 57: Similarly, it would be better if we have access to a constructor similar to `EKI.jl` (not saying this one should be deleted).

examples/Lorenz63/Lorenz63.jl
```diff
@@ -0,0 +1,188 @@
+using JLD2
```
ilopezgp, line 1: JLD might still be a better idea than JLD2. See from the JLD2 GitHub: _Please use caution. If your tolerance for data loss is low, JLD may be a better choice at this time._

It also offers more compatibility with the HDF5 format, so we can read results in Python for example.

odunbar: Initially I had no intentions to set a precedent with this. So long as we can store the objects we desire then I don't mind which we use.  Though this is consistent with the current codebase - so perhaps raise an issue and this can be resolved in another PR.

odunbar:
@Zhengyu-Huang After our brief chat the other day I think we would like both of the following in the `EnsembleKalmanProcesses` package. that is, 
1) the Lorenz 63 example
2) the UKI updating method for EKP.
What I would recommend is to basically first move this PR to `CliMA/EnsembleKalmanProcesses.jl`. Then I'd be happy to chat further about the framework to actually get it to fit the EKP format.

odunbar:
Closing for now due to move of PR to `EnsembleKalmanProcesses.jl`

Issue #73, opened 2020-08-05
Avoid unnecessary matrix inversions


ilopezgp:
There are several instances of matrix inversion in the source code where it is only necessary to perform a linear solve. 

Matrix inversion is more expensive and less accurate than linear solvers, so all these instances should be deleted.

PR #74, opened 2020-08-05
Substitute unnecessary matrix inversion, use linear solve instead.

ilopezgp:
This PR solves issue #73 . Inverting the observational covariance matrix is not necessary, a linear solve can be used instead. The linear solve is faster and more accurate, which can be important when the observational covariance matrix is large or ill-conditioned.

### DIFF
--- a/src/MCMC.jl
+++ b/src/MCMC.jl
@@ -33,8 +33,6 @@ struct MCMCObj{FT<:AbstractFloat, IT<:Int}
     obs_sample::Vector{FT}
     "covariance of the observational noise"
     obs_noise_cov::Array{FT, 2}
-    "inverse of obs_noise_cov"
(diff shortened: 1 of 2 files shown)


ilopezgp:
@odunbar I ran the `test/MCMC/runtests.jl` with this version of the code. It takes an average of 12.8 s to run in my computer. I ran the same test with master and it took an average of 13.5 s to run. At least for the MCMC test, this is faster. This PR always uses a linear solve with a diagonal matrix (stored as a `Diagonal` type), so the linear solve seems to be quick enough to not slow down the code.

odunbar:
Yeah I expected if the type was recognized, then it would hopefully do something more efficient! Good to know. I'm happy for this to be merged

PR #76, opened 2020-08-11
Change code infrastructure to accommodate both EKI and EKS

bielim:
This PR is an interim solution, to be superseded by PR #70 as soon as the EKS implementation has been finalized. It implements the new "EK infrastructure" with the abstract type `Process` that allows the user to choose between an EK `Sampler` and an EK `Inversion`, but since the correct ensemble update for the EK Sampler is not known yet (see discussion in the comment section of PR #70), an attempt to instantiate an EK `Sampler` is prohibited and will throw an error. 
The point of merging this PR before completion of the EK Sampler implementation is to not block the further development of the code base, which depends on the infrastructure introduced in this PR.


bielim:
bors try

odunbar:
Thanks Melanie - This is exactly what we need in the meantime. Lets hope the update will be resolved soon! I think this is a good lesson to us all about why we should make small targeted PRs that can be merged quickly to reduce dependency :)

PR #79, opened 2020-10-29
add mindmap link to README.md

odunbar:


README.md
```diff
@@ -15,3 +15,7 @@
 [bors-img]: https://bors.tech/images/badge_small.svg
 [bors-url]: https://app.bors.tech/repositories/24774
 
+
+### Project Mind map
+The latest code structure is found at:
+https://miro.com/welcomeonboard/j4n8NPSihwQ3X7wjEH7kkitUi6I9deUa5veFV15dCndrTNRZkczGGyIviVjnuAhZ
```
ilopezgp, line 21: Does this link allow anyone to modify the mind map? Since this repo is open source, we should probably limit the functionality to a number of users.

odunbar: This is a editable link. I think we either have a read-only option otherwise. I think I am happy to have this open for edits as it stands, but later we can close it off?

odunbar: PS it's been waiting for bors for like 5 days...

ilopezgp: We can leave it open, I don't think there will be any problems for now. Is there version control though? In case someone tries to delete/modify everything?

ilopezgp: Here is a link discussing how the owner can set up roles and **save a board backup**: https://help.miro.com/hc/en-us/articles/360017571194-Roles-in-Miro . We probably want to do the latter at least.

PR #83, opened 2020-11-03
Corrected EKS implementation

agarbuno:
This kills a bug in the EKS method and add a linear example as a test.


Project.toml
```diff
@@ -9,6 +9,7 @@ Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 GaussianProcesses = "891a1506-143c-57d2-908e-e1f8e92e6de9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
```
ilopezgp, line 12: Do we want to keep this dependency on Plots or do something else? @jakebolewski @bielim @odunbar

odunbar: I think a good thing to do with plots (for this PR) is described here https://discourse.julialang.org/t/add-package-to-project-toml-when-needed-only-for-testing/20552/3 
In our setting I think it amounts to the following

1. Add `Test` to the Project.toml
2. Add (manually) new sections `[extras]` and `[targets]`  into Project.toml,
3. Copy the `Test` line and the `Plots` lines into `[extras]`
4. in the `[targets]` section put `test =["Test","Plots"]`

Then when you call Pkg.test("MyPackage"), (not run runtest.jl directly) it will also load the Plots package.

NB I haven't done this before but hopefully it will work

src/EKP.jl
```diff
@@ -74,11 +74,11 @@ function EKObj(parameters::Array{FT, 2},
 
     # Throw an error if an attempt is made to instantiate a `Sampler` EKObj
     # EK Sampler implementation is not finalized yet, so its use is prohibited
-    # TODO: Finalize EKS implementation (can be done as soon as we know the 
+    # TODO: Finalize EKS implementation (can be done as soon as we know the
```
odunbar, line 77: remove these comments?

src/EKP.jl
```diff
@@ -74,11 +74,11 @@ function EKObj(parameters::Array{FT, 2},
...
     #       correct EKS update equation, which apparently is different from
-    #       Eq. (2.8) in Cleary et al. (2019)) 
-    err_msg = "Ensemble Kalman Sampler is not fully implemented yet. Use Ensemble Kalman Inversion instead."
-    typeof(process) != Sampler{FT} || error(err_msg)
+    #       Eq. (2.8) in Cleary et al. (2019))
+    # err_msg = "Ensemble Kalman Sampler is not fully implemented yet. Use Ensemble Kalman Inversion instead."
```
odunbar, line 80: remove these comments?

src/EKP.jl
```diff
@@ -246,22 +246,24 @@ function update_ensemble!(ek::EKObj{FT, IT, Sampler{FT}}, g) where {FT, IT}
...
+    D = (1/N_ens) * (E' * (ek.obs_noise_cov \ R))
 
     Δt = 1/(norm(D) + 1e-8)
+
     noise = MvNormal(u_cov)
 
+
     ###########################################################################
     ###############    TODO: Implement correct equation here   ################
     ###########################################################################
```
odunbar, line 258: remove comment

src/EKP.jl
```diff
@@ -234,8 +234,8 @@ function update_ensemble!(ek::EKObj{FT, IT, Sampler{FT}}, g) where {FT, IT}
     N_ens = ek.N_ens
 
     # u_mean: N_params x 1
-    u_mean = mean(u', dims=2) 
-    # g_mean: N_params x 1 
+    u_mean = mean(u', dims=2)
+    # g_mean: N_params x 1
     g_mean = mean(g', dims=2)
     # g_cov: N_params x N_params
     g_cov = cov(g, corrected=false)
```
odunbar, line 241: corrected=false is a default and can be removed

test/EKP/runtests-linear-model.jl
```diff
@@ -0,0 +1,137 @@
...
+
+    posterior_cov_inv = (A' * (Γy\A) + 1 * Matrix(I, n_par, n_par)/prior_cov)
+    ols_mean          = (A'*(Γy\A)) \ (A'*(Γy\y_obs))
+    posterior_mean    = posterior_cov_inv \ ( (A'*(Γy\A)) * ols_mean  + (prior_cov\prior_mean))
+
+    #### This tests correspond to:
+    # EKI provides a solution closer to the ordinary Least Squares estimate
+    @test norm(ols_mean - eki_final_result) < norm(ols_mean - eks_final_result)
+    # EKS provides a solution closer to the posterior mean
+    @test norm(posterior_mean - eks_final_result) < norm(posterior_mean - eki_final_result)
```
odunbar, line 130: Are these tests always going to be true? Or just in this test case, e.g this depends on realizations of the data you provide etc?

test/EKP/runtests-linear-model.jl
```diff
@@ -0,0 +1,137 @@
...
+
+    ##### I expect this test to make sense:
+    # In words: the ensemble covariance is still a bit ill-dispersed since the
+    # algorithm employed still does not include the correction term for finite-sized
+    # ensembles.
+    @test abs(sum(diag(posterior_cov_inv\cov(eksobj.u[end]))) - n_par) > 1e-5
```
odunbar, line 136: This is maybe more of a feature of the current algorithm as opposed to a test. e.g are we actually implying we fail this test if the covariance is not ill-dispersed?

Project.toml
bielim: Let's try this, yes (this is what worked for @glwagner in Oceananigans, right?)

bielim (commented):
I suggest that we make `runtests-linear-model.jl` the new `runtests.jl` and delete the old one, since `runtests-linear-model.jl` is basically an improved version of the example in `runtests.jl`. But it needs some more @test statements - in general it would be good if all functions, objects and object attributes in `EKP.jl` were "touched" at least once in these tests.

src/EKP.jl
bielim: According to the [docs](https://docs.julialang.org/en/v1/stdlib/Statistics/), the default is actually `corrected=true`, so we need to explicitly set the flag to `false` here.

test/EKP/runtests-linear-model.jl
bielim: As discussed, we will improve the tests in a future PR, this one is mainly about implementing the EKS.

test/EKP/runtests-linear-model.jl
bielim: As discussed, we will improve the tests in a future PR, this one is mainly about implementing the EKS.

PR #84, opened 2020-11-12
Merge duplicate Ag/eks_imlplementation into original ag/eks-implementation (with new commits)

odunbar:


odunbar:
I will merge without review as it is clear what has been done.

PR #89, opened 2020-12-08
WIP: replace Prior and posterior samples with ParameterDistributions

odunbar:
# Purpose
 
To follow from PR #88 in replacing the prior distributions and posterior distributions with the new type ParameterDistributions, and adding the requisite functionality to make this possible.

## Contained in the PR

- Implement methods: `get_logpdf`, `get_cov`,`get_mean` and replace implementation in EKP, and MCMC. Note this will also allow us to use prior distributions with block diagonal (i.e not only diagonal) in the MCMC. 
- Add requisite unit tests 
- Modify `runtests.jl`  that are dependent on `Priors.jl`, to instead use ParameterDistributions
- Remove Priors.jl

**Future PR will deal with example cases (not contained in runtests)**

## Additionally
- [x]  Created the following issue: When creating EKS, before one supplied mean and cov separately, these can now be deduced from the prior (which is also an input).


src/ParameterDistribution.jl
```diff
@@ -335,7 +343,24 @@ function get_cov(pd::ParameterDistribution)
     return cat(block_cov...,dims=(1,2)) #build the block diagonal (dense) matrix
     
 end
-    
+"""
+    get_mean(pd::ParameterDistribution)
+
+returns a mean of the distirbutions
+"""
+
```
charleskawczynski, line 351:
```suggestion
```

odunbar: Accident?

odunbar:
Recieved a review offline from @bielim . Merging

PR #92, opened 2020-12-18
[WIP] Update examples to work with latest CES code

bielim:
The goal of this PR is to get all examples synced up with the latest changes in the code base (in particular, PRs  #88 and #89)

- [x]  `Cloudy_example.jl`
- [x]  `learn_noise.jl`
- [x] `plot_GP.jl`

In addition, `get_distribution()` (in `ParameterDistribution.jl`) has been modified to return the array of samples when called for `Samples` ( rather than the message "Contains samples only"). `get_distribution` now returns a `Dict` with the parameter names as keys and the corresponding distribution (in the case of `Parameterized` distributions, such as Normal(0.0, 1.0)) or the corresponding samples (in the case of parameters represented by `Samples`) as a parameter_dimension x n_samples array.


examples/Cloudy/Cloudy_example.jl
```diff
@@ -216,42 +253,40 @@ mcmc = MCMC.MCMCObj(yt_sample, Γy, priors, new_step, u0, max_iter,
...
     else
         throw("not implemented")
     end
 
-    label = "true " * param
-    histogram(posterior[:, idx], bins=100, normed=true, fill=:slategray, 
+    label = "true " * names[idx]
+    println("size of samples to be plotted in histogram")
+    println(size(posterior.distributions[idx].distribution_samples))
+    ys = dropdims(posterior.distributions[idx].distribution_samples, dims=1)
```
odunbar, line 282: Could you use the proper getter method here? e.g `get_n_samples(posterior)` returns a dictionary for each distribution, if they are of `Samples` type it will have the number of samples.

examples/Cloudy/Cloudy_example.jl
```diff
@@ -216,42 +253,40 @@ mcmc = MCMC.MCMCObj(yt_sample, Γy, priors, new_step, u0, max_iter,
...
 println("post_cov")
 println(post_cov)
-println("D util")
-println(det(inv(post_cov)))
-println(" ")
 
 # Plot the posteriors together with the priors and the true parameter values
-true_values = [log(N0_true) log(θ_true) log(k_true)]
-n_params = length(true_values)
+n_params = length(names)
```
odunbar, line 266: Just a note, you could use a getter here from the parameter distribution `get_total_dimension(posterior)` to obtain this - might be nice to showcase here

examples/Cloudy/Cloudy_example.jl
```diff
@@ -216,42 +253,40 @@ mcmc = MCMC.MCMCObj(yt_sample, Γy, priors, new_step, u0, max_iter,
...
+    label = "true " * names[idx]
+    println("size of samples to be plotted in histogram")
+    println(size(posterior.distributions[idx].distribution_samples))
+    ys = dropdims(posterior.distributions[idx].distribution_samples, dims=1)
+    histogram(ys, bins=100, normed=true, fill=:slategray, thickness_scaling=2.0,
               lab="posterior")
-    plot!(xs, mcmc.prior[idx].dist, w=2.6, color=:blue, lab="prior")
-    plot!([true_values[idx]], seriestype="vline", w=2.6, lab=label)
+    println("created histogram")
+    plot!(xs, get_distribution(mcmc.prior.distributions[idx]), w=2.6,
```
odunbar, line 286: Could we possibly have this as `get_distribution(mcmc.prior)` which provides a dictionary of the distributions - we then plot from the dictionary?

odunbar (commented):
Looks good! I think the heart of my comments are that we have built an interface with `ParameterDistribution` s, and so we should use always use it. That is, rather than directly accessing the (effectively private) members e.g `posterior.distribution[i]` we should be instead "getting" them. i.e call `get_distribution(posterior)` and then using the returned objects. This is typically easier to maintain down the line if everything interacts with the distributions in the same way. If the right getter functionality isn't there we should add the getter functions over bypassing them.

examples/Cloudy/Cloudy_example.jl
bielim: These were debugging comments that have been removed.

examples/Cloudy/Cloudy_example.jl
bielim: I changed this to `length(get_name(posterior))` (What I need here is the number of parameters, `get_total_dimension(posterior)` would only return the correct number if all parameters are 1-D)

examples/Cloudy/Cloudy_example.jl
bielim: Yep, done! I was thinking that for consistency it would be nice if `get_distribution()` also returned a dictionary (rather than the message "Contains samples only") for `Samples`. In the case of `Samples`, the dictionary's values could either be the `Samples` object itself or the samples in the form of an array. 
This would remove the inconsistency in how the prior and the posterior distribution are accessed here - the former is accessed using `get_distribution`, the latter via accessing the member itself (`posterior.distributions[idx].distribution_samples`)

bielim: Implemented the change to `get_distribution` suggested above.

bielim:
Superseded by #94 (messed up the squash and rebase :-))

Issue #82, opened 2020-11-02
Use `EnsembleKalmanProcess` rather than `EKObj`?


glwagner:
https://github.com/CliMA/CalibrateEmulateSample.jl/blob/b210ee85f4a4fe9be4223cd8e80f29f6ce7faaeb/src/EKP.jl#L46

`EnsembleKalmanProcess` may be more literal than `EKObj`. Also, using a verbose form for the name will make scripts more readable, especially for non-experts (and a lot of non-statistologists physicsts may read these scripts, considering their application).

cc @odunbar

glwagner:
Likewise, the module `EKP.jl` could be called `EnsembleKalmanProcesses.jl`.

ali-ramadhan:
I second this suggestion. Since it's a breaking change, it would probably be good to do this before tagging a v0.1.0.

odunbar:
Because you asked so nicely, I'm working on a branch to deal with these. `orad/verbose-names` 

I will include filenames and the objects within.
`EKP.jl` -> `EnsembleKalmanProcesses.jl`
`GPEmulator.jl` -> `GaussianProcessEmulator.jl`
`MCMC.jl` -> `MarkovChainMonteCarlo.jl`

PR #95, opened 2020-12-23
More verbose naming

odunbar:
## Purpose
To introduce more verbose naming conventions for filenames, modules and structs in the interest of more verbose naming conventions, as raised in Issue #82 

## Content
Name changes:
 - `GPEmulator` to `GaussianProcessEmulator` and `GPObj` to `GaussianProcess`
 - `EKP` to `EnsembleKalmanProcesses` and `EKObj` to `EnsembleKalmanProcess`
 - `MCMC` to `MarkovChainMonteCarlo` and `MCMCObj` to `MCMC` 
 - find_ek_step -> find_ekp_stepsize
 - extract_GP_tp -> get_training_points

Changes made throughout: 
- [x] src
- [x] docs
- [x] tests
- [x] examples  

### Additional tests
It seems the current runtests for Utilities and EKP are below the required codecov standard, therefore I shall write some additional tests before the PR can be merged.

Coauthored with @ilopezgp


ilopezgp:
When running `test/EnsembleKalmanProcesses/runtests.jl`, it seems like `TEST_PLOT_OUTPUT` is not defined. Can you verify this behavior @odunbar ?

test/EnsembleKalmanProcesses/runtests.jl
```diff
@@ -77,6 +82,7 @@ using CalibrateEmulateSample.ParameterDistributionStorage
     # Plot evolution of the EKI particles
     eki_final_result = vec(mean(ekiobj.u[end], dims=1))
     
+    TEST_PLOT_OUTPUT = false
```
odunbar, line 85: > When running test/EnsembleKalmanProcesses/runtests.jl, it seems like TEST_PLOT_OUTPUT is not defined. Can you verify this behavior @odunbar ?

In theory we should run the tests as a block i.e from `test/runtests.jl` and so this is where TEST_PLOT_OUTPUT is defined (via a get method with 0 default if the CES_TEST_PLOT_OUTPUT variable is not defined). If you directly run the EKP block from `test/EnsembleKalmanProcesses/runtests.jl` you would need to define it. 

So i would remove the line you've added here. Otherwise these run for me! Thanks!

odunbar: Edit: I've removed this in latest commit.

ilopezgp: Great, thanks!

ilopezgp (approved):

PR #96, opened 2020-12-24
Adds LossMinimization folder and two examples

sandreza:
This is meant to be a starting point for a discussion on adding tutorials on how to use calibrate emulate sample with simple examples to build intuition.

In particular PR adds a simple example showing how to use EKI to minimize two loss functions
1. L(x,y) = (x-1)^2 + (y+1)^2
2. L(x,y) = (x^2-1)^2 + (y+1)^2

This is accomplished by taking the forward map to just be the loss function. The former has a minimum at (x,y) = (1,-1) and the latter has two minima at (x,y) = (+1,-1) and (x,y) = (-1,-1). By switching the random seed on the latter, one can converge to either minimum.  Eventually, such simple examples could be included in the docs

### DIFF
--- a/examples/LossMinimization/loss_minimization.jl
+++ b/examples/LossMinimization/loss_minimization.jl
@@ -1,0 +1,117 @@
+using Distributions
+using LinearAlgebra
+using Random
+using Test
+using Plots
+using CalibrateEmulateSample.EKP
+using CalibrateEmulateSample.ParameterDistributionStorage
+# Seed for pseudo-random number generator for reproducibility
+rng_seed = 41
+Random.seed!(rng_seed)
+
+# Number of synthetic observations from G(u)
+n_obs = 1
+# Defining the observation noise level
+noise_level =  1e-8   
+# Independent noise for synthetic observations       
+Γy = noise_level * Matrix(I, n_obs, n_obs) 
+noise = MvNormal(zeros(n_obs), Γy)
+
+# Loss Function (unique minimum)
+function G(u)
+    return [sqrt((u[1]-1)^2 + (u[2]+1)^2)]
+end
+
+# Loss Function Minimum
+u_star = [1.0, -1.0]
+y_obs  = G(u_star) + 0 * rand(noise) 
+
+# Define Prior
+prior_distns = [Parameterized(Normal(0., sqrt(1))),
+                Parameterized(Normal(-0., sqrt(1)))]
+constraints = [[no_constraint()], [no_constraint()]]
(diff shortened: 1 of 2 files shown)


odunbar:
Thanks for starting this! I'd be keen to chat more about some tutorial-type examples. At present we only have the `EnsembleKalmanProcesses/runtests` (`EKP/runtests`) for the ensemble methods (in isolation).

sandreza:
"we can relegate the runtests to more unit-style function testing"
I agree with this!

examples/LossMinimization/loss_minimization.jl
```diff
@@ -0,0 +1,117 @@
+using Distributions
+using LinearAlgebra
+using Random
+using Test
```
odunbar, line 4: Test is not used

examples/LossMinimization/loss_minimization.jl
```diff
@@ -0,0 +1,117 @@
...
+using Plots
+using CalibrateEmulateSample.EKP
```
odunbar, line 6: Renamed to `CalibrateEmulateSample.EnsembleKalmanProcesses` in newest master

examples/LossMinimization/loss_minimization.jl
```diff
@@ -0,0 +1,117 @@
...
+prior_mean = reshape(get_mean(prior),:)
+prior_cov = get_cov(prior)
+
+# Calibrate
+N_ens = 50  # number of ensemble members
+N_iter = 20 # number of EKI iterations
+initial_ensemble = EKP.construct_initial_ensemble(prior, N_ens;
+                                                rng_seed=rng_seed)
+
+ekiobj = EKP.EKObj(initial_ensemble,
```
odunbar, line 44: `EKP` -> `EnsembleKalmanProcesses`
`EKObj` -> `EnsembleKalmanProcess`

examples/LossMinimization/loss_minimization.jl
sandreza: removed

examples/LossMinimization/loss_minimization.jl
sandreza: and renamed!

examples/LossMinimization/loss_minimization.jl
sandreza: also renamed

odunbar:
bors try

odunbar (approved):

Issue #93, opened 2020-12-21
Incorrect instantiation of posterior `ParameterDistribution` in `get_posterior` 


bielim:
In `get_posterior` (in `MCMC.jl`), a `ParameterDistribution` gets created with the posterior samples. The posterior samples are an mxn array, which gets interpreted by the `Samples` object as m samples of an n-dim parameter. However, the posterior mxn array may actually represent several parameters whose dimensions sum up to n rather than a single n-dim parameter. In these cases instantiating the ParameterDistribution will cause `DimensionMismatch` errors.

PR #101, opened 2021-01-20
Adding L96 example

mhowlan3:
Addresses Issue #99 
Also Addresses Issue #93 
Adds a L96 example. Has a "seasonal" forcing option. RK4 time integration.


examples/Lorenz/Lorenz_example.jl
```diff
@@ -0,0 +1,381 @@
...
+using JLD2
+
+# Import Calibrate-Emulate-Sample modules
+using CalibrateEmulateSample.EnsembleKalmanProcesses
+using CalibrateEmulateSample.GaussianProcessEmulator
+using CalibrateEmulateSample.MarkovChainMonteCarlo
+using CalibrateEmulateSample.Observations
+using CalibrateEmulateSample.Utilities
+using CalibrateEmulateSample.ParameterDistributionStorage
+
```
odunbar, line 21: Perhaps could you write a short doc-string to explain briefly what is done in the script.
- what the model is
- what parameters you learn 
- what data is used 
- what is the output of the file

examples/Lorenz/Lorenz_example.jl
```diff
@@ -0,0 +1,381 @@
...
+rng_seed = 4137
+Random.seed!(rng_seed)
+
+# Output figure save directory
+figure_save_directory = "/home/mhowland/Codes/CESPlots/figures/lorenz_96_dynamic/debug/T_180/"
+data_save_directory = "/home/mhowland/Codes/CESPlots/data/lorenz_96_dynamic/debug/T_180/"
+
+# Governing settings
```
odunbar, line 29: I think we want to reorganize slightly, Basically to separate the forward model from the parameter learning.
- Setup only that which is required to run the "truth", e.g all numerical parameters and "true" parameter values
- Run the truth run
- Then, set up the priors for recovering parameters of choice and initialize EKI etc. 
- initialize EKI ...

examples/Lorenz/Lorenz_example.jl
```diff
@@ -0,0 +1,381 @@
...
+# Characteristic time scale
+τc = 5. # days, prescribed by the L96 problem
+# Stationary or transient dynamics
+dynamics = 2 # Transient is 2
+# Lognormal prior or normal prior?
+log_normal = false # THIS ISN't CURRENTLY IMPLEMENTED
```
odunbar, line 35: do we want the option if not implemented? or would you prefer to remove it

odunbar (changes requested):
Looks great! Overall we need 2 main changes (1) move the Project/Manifest changes out to the Lorenz directory (2) separate the parameters/settings for the model from the parameters/setting from the learning setup!

Thanks!

examples/Lorenz/Lorenz_example.jl
mhowlan3: It should be a bit better now but I'm open to future modifications.

mhowlan3: It should be a bit better now but I'm open to future modifications.

examples/Lorenz/Lorenz_example.jl
mhowlan3: It should be a bit better now but I'm open to future modifications.

examples/Lorenz/Lorenz_example.jl
mhowlan3: It could be something someone wants to implement later, and there are flags for plotting.

examples/Lorenz/Lorenz_example.jl
```diff
@@ -0,0 +1,398 @@
+# Import modules
+include("GModel.jl") # Contains Lorenz 96 source code
+
+include("/groups/esm/mhowland/Codes/CalibrateEmulateSample.jl/src/ParameterDistribution.jl")
```
odunbar, line 4: For testing, add `CalibrateEmulateSample` into the package manager by calling
`develop /groups/esm/mhowland/Codes/CalibrateEmulateSample.jl`
then you can use the commented statements below and remove these local includes

mhowlan3: Done, thanks!

examples/Lorenz/Lorenz_example.jl
mhowlan3: Yes, I will do this.

odunbar (approved):
LGTM! Thanks a lot Mike

PR #109, opened 2021-03-11
add buildkite setup to build examples

jakebolewski:
Example build output: https://buildkite.com/clima/calibrateemulatesample-ci/builds/39

Right now it is not hooked up to bors (so passing example build is optional) but we can change that if desired.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,102 @@
+env:
+  JULIA_VERSION: "1.5.2"
+  OPENBLAS_NUM_THREADS: 1
+  GKSwstype: nul
+
+steps:
+  - label: "init project env"
+    key: "init_project_env"
+    command:
+      - "echo $$JULIA_DEPOT_PATH"
+      - "julia --project -e 'using Pkg; Pkg.instantiate(;verbose=true)'"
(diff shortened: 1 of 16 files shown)


jakebolewski:
bors try

odunbar (approved):
LGTM - outputs are the same on my laptop.

Can you add to the wiki examples section something telling users what to add regarding these load paths, so that figures work on the online setup?
e.g lines such as
```Julia
prepend!(LOAD_PATH, [joinpath(@__DIR__, "..", "..")])
include(joinpath(@__DIR__, "..", "ci", "linkfig.jl"))
```

jakebolewski:
the link figure bit is only for ci, it's basically a no-op if it doesn't detect the buildkite env to display figures in the logs.

the load path here just references the in tree version of whatever the working version of CalibrateEmulateSample is in the repo, so if you change CES and then re-run the examples you will see the examples reflect those in tree changes. 

The practical implication of this is you have to first build the CES project before running the examples: 

```
cd CalibrateEmulateSample.jl
julia --project -e 'using Pkg; Pkg.instantiate()'
#... possibly make changes...
julia --project=examples/Lorenz -e 'using Pkg; Pkg.instantiate()'
julia --project=examples/Lorenz  examples/Lorenz/Lorenz_example.jl
```

I can add these steps to the README

PR #111, opened 2021-03-23
Regularization & standardization

mhowlan3:
1. Log-posterior calculation in MCMC
2. SVD regularization via truncation
3. Re-normalization (standardization) of output space


src/GaussianProcessEmulator.jl
```diff
@@ -457,12 +462,32 @@ Note: If F::SVD is the factorization object, U, S, V and Vt can be obtained via
...
-        sqrt_singular_values_inv = Diagonal(1.0 ./ sqrt.(decomposition.S)) 
-        transformed_data = sqrt_singular_values_inv * decomposition.Vt * data
+        # MFH, 3/22/21: Truncate the SVD as a form of regularization
+	if truncate_svd # this variable needs to be provided to this function
+            # Perform SVD
+            decomposition = svd(obs_noise_cov)
+            # Find cutoff
+            σ = decomposition.S
+            σ_cumsum = cumsum(σ) / sum(σ);
+            P_cutoff = 0.95;
```
odunbar, line 475: `P_cutoff` should be allowed to be changed by the user.  
`svdvals(...)` is maybe neater than `svd(...)` for the first case, as all we want is the singular values to calculate the information content. (though I expect it is very similar performance-wise)

src/GaussianProcessEmulator.jl
```diff
@@ -472,17 +497,38 @@ function svd_transform(data::Array{FT, 2}, obs_noise_cov::Union{Array{FT, 2}, No
 end
 
 function svd_transform(data::Vector{FT}, obs_noise_cov::Union{Array{FT, 2}, Nothing}) where {FT}
```
odunbar, line 499: We can define the svd_transform function outside of the GP constructor, so we don't need a copy of it in here

src/MarkovChainMonteCarlo.jl
```diff
@@ -173,8 +183,15 @@ function log_likelihood(mcmc::MCMC{FT},
...
-        log_gpfidelity = -FT(0.5) * log(det(Diagonal(gvar))) # = -0.5 * sum(log.(gvar))
-        diff = g - mcmc.obs_sample
+	# det(log(Γ))
+	# Ill-posed numerically for ill-conditioned covariance matrices with det≈0
+        #log_gpfidelity = -FT(0.5) * log(det(Diagonal(gvar))) # = -0.5 * sum(log.(gvar))
+	# Well-posed numerically for ill-conditioned covariance matrices with det≈0
+	full_cov = Diagonal(gvar)
+	eigs = eigvals(full_cov)
+	log_gpfidelity = -FT(0.5) * sum(log.(eigs))
+	# Combine got log_rho
```
odunbar, line 193: In this version of MCMC, we are actually assuming that `gvar` is a vector of floats, I'm thinking maybe we should relax this to take in matrices. Perhaps the easiest way is to rewrite the current `mcmc_sample!` function to take in matrix input `gcov` and add this extra function
```Julia 
function mcmc_sample!(mcmc::MCMC{FT}, g::Vector{FT}, gvar::Vector{FT}) where {FT}
    return mcmc_sample!(mcmc, g, Diagonal(gvar))
end
```

odunbar: I've just checked that if eigvals receives a  `Diagonal` type matrix it really does just take the diagonal as the eigenvalues. and so is a nanosecond timed computation still.

src/GaussianProcessEmulator.jl
```diff
@@ -515,4 +588,12 @@ function svd_reverse_transform_mean_cov(μ::Array{FT, 2}, σ2::Array{FT, 2}, dec
     return transformed_μ, transformed_σ2
 end
 
+function get_standardizing_factors(data::Array{FT,2}) where {FT}
```
odunbar, line 591: This is maybe best put into `src/Utilities.jl`? There are some z-score type normalizations there too.

mhowlan3: Good idea, it's moved now. This function is not general anyway, users really would need to supply this.

src/MarkovChainMonteCarlo.jl
mhowlan3: Great!

src/GaussianProcessEmulator.jl
mhowlan3: Done!

examples/Lorenz/Project.toml
```diff
@@ -6,6 +6,7 @@ DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 GaussianProcesses = "891a1506-143c-57d2-908e-e1f8e92e6de9"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
odunbar, line 9: You should also remove DifferentialEquations from the Project.toml as we don't use it any more

src/MarkovChainMonteCarlo.jl
odunbar: I am not sure you have pushed this yet? I can't see it it

src/GaussianProcessEmulator.jl
```diff
@@ -66,11 +76,13 @@ struct GaussianProcess{FT<:AbstractFloat, GPM}
     "the Gaussian Process (GP) Regression model(s) that are fitted to the given input-data pairs"
     models::Vector
     "the singular value decomposition of obs_noise_cov, such that obs_noise_cov = decomposition.U * Diagonal(decomposition.S) * decomposition.Vt."
-    decomposition::Union{SVD, Nothing}
+    decomposition::Union{SVD, decomp_struct, Nothing}
```
odunbar, line 79: It's fine to create the decomposition structure here, 
might be nice to save everything in you `decomp_struct` format, then you can use `Union{decomp_struct, Nothing}` as the type.

You just need to add a constructor for `decomp_struct` for example
```julia
decomp_struct(svd::SVD) = decomp_struct(svd.V, svd.Vt, svd.S, size(svd.V)[0])
```

odunbar (approved):
Thanks very much Mike! 
1. If you agree about the decomposition type, could you change that, should be a quick change
2. Check that you wish to keep all the `println` statements in the source files, and also there are a couple of `# TO DO` comments that could be removed in the source files (check the diff)

Otherwise LGTM!

PR #112, opened 2021-03-23
build and deploy docs

jakebolewski:


odunbar (approved):
LGTM.

I like having minimal doc dependencies, and thanks for the docs initial installation instructions.

Issue #121, opened 2021-12-02
Unit tests broken under Julia 1.7 due to RNG change


tsj5:
## Problem description

Running unit tests on a fresh installation of the `master` branch results in 3 failures: two in `GaussianProcessEmulator`, one in `Utilities`.

## Steps to reproduce

Under `arm64` julia 1.7.0 on macOS 12.0.1, `julia --project -e 'using Pkg; Pkg.test()'` gives

<details>
  <summary>CLI output</summary>

(package dependency output omitted)

```
     Testing Running tests...
Starting tests for GaussianProcessEmulator
Using user-defined kernelType: SEIso{Float64}, Params: [0.0, 0.0]
Learning additive white noise
kernel in GaussianProcess
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [0.0, 0.0]  Type: Noise{Float64}, Params: [0.0]
created GP1
optimized GP1
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [0.4671112501592054, -0.11637219100301877]  Type: Noise{Float64}, Params: [-2.9126145297418646]
Using user-defined kernelType: SEIso{Float64}, Params: [0.0, 0.0]
Learning additive white noise
kernel in GaussianProcess
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [0.0, 0.0]  Type: Noise{Float64}, Params: [0.0]
created GP1
optimized GP1
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [-0.055491995428277534, -0.11637219133420702]  Type: Noise{Float64}, Params: [-2.912614529547923]
Using user-defined kernelType: SEIso{Float64}, Params: [0.0, 0.0]
Learning additive white noise
kernel in GaussianProcess
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [0.0, 0.0]  Type: Noise{Float64}, Params: [0.0]
created GP1
optimized GP1
Type: SumKernel{SEIso{Float64}, Noise{Float64}}
  Type: SEIso{Float64}, Params: [0.4671112501592054, -0.11637219100301877]  Type: Noise{Float64}, Params: [-2.9126145297418646]
[...]

Issue #81, opened 2020-11-02
Docs mention module EKI.jl but there's only EKP.jl


glwagner:
![image](https://user-images.githubusercontent.com/15271942/97878203-517df180-1cec-11eb-849a-e42b376683fc.png)

odunbar:
Resolved in PR #139

PR #122, opened 2021-12-05
[WIP] Use EKP.jl 0.2.0 & pass explicit RNG object where used

tsj5:
[*Edited 1 Feb 2022 in response to updates*] 

This PR updates CES to use the [0.2.0 release](https://github.com/CliMA/EnsembleKalmanProcesses.jl/releases/tag/v0.2.0) of EnsembleKalmanProcesses.jl (EKP). In addition to using changed module and method names, the PR adds support in CES for use of an explicit random number generator where needed; this was done in EKP [#81](https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/81).

[Recommended practice](https://discourse.julialang.org/t/set-the-state-of-the-global-rng/12599/3) for reproducible tests is to pass an RNG object explicitly to all methods which need it. It's not sufficient to set the global seed for the default RNG: the default RNG algorithm used by Julia was [changed](https://julialang.org/blog/2021/11/julia-1.7-highlights/#new_rng_reproducible_rng_in_tasks) in version 1.7. This will affect us in the future when we update CES to support Julia > 1.6 (CliMA/CalibrateEmulateSample.jl#121): the [docs](https://docs.julialang.org/en/v1/stdlib/Random/#Reproducibility) for the Random package note that the same seed may give different random values in different versions.

Changes made to the CES API in this PR are minor (which is why the RNG change is combined with the EKP update):
- An `rng` field is added to `MCMC`; the constructor sets this to `Random.GLOBAL_RNG` by default.
- `rng` is recognized as an optional first argument to `get_obs_sample`, used in lieu of the `rng_seed` kwarg if provided.

All changes are fully backwards compatible. In particular, tests are updated to use an explicit rng, but examples have been left unchanged. It may be desirable to update them as well.


odunbar (approved):
A couple of small things,
- could you rename the PR to include the updating to the new EKP. 
- I guess we should also add `EnsembleKalmanProcesses="^0.2" to the [compat]` in the `Project.toml`
- is the intention to include the MCMC scratch in the PR? is it used anywhere/useful? If so could you make that more clear? 

Otherwise good to see this passes the tests with the new EKP! LGTM.

tsj5:
> A couple of small things,

Done -- thanks for catching the accidentally committed scratch file; I've moved my scratch work out of the repo to prevent mistakes like this in the future.

odunbar:
LGTM ! Merge away

tsj5:
> Build failed: [buildkite/calibrateemulatesample-ci](https://buildkite.com/clima/calibrateemulatesample-ci/builds/61)

Only failure is in the Cloudy example, as described in #131. I should spend more time banging my head against that issue; right now I don't have a good understanding of the conditions needed to reproduce it (i.e. why the bug is present on this branch and not `master`).

tsj5:
The test-macos failure ([Test \#307](https://github.com/CliMA/CalibrateEmulateSample.jl/runs/5203658428?check_suite_focus=true)) above came from bors running the Github Actions workflow, which was a puzzle since the same test passed immediately beforehand when run directly ([test \#306](https://github.com/CliMA/CalibrateEmulateSample.jl/runs/5194650850?check_suite_focus=true)). The error happened in [line 50](https://github.com/tsj5/CalibrateEmulateSample.jl/blob/bf4c2e21e4d12d82ae8acbf1ff001bbd80daf2a6/test/MarkovChainMonteCarlo/runtests.jl#L50) of the MCMC tests, while fitting the first Gaussian process:
```
2022-02-15T16:43:28.2088540Z created GP: 1
2022-02-15T16:43:28.4049700Z PosDefException(37)
2022-02-15T16:43:34.9495550Z MarkovChainMonteCarlo: Error During Test at /Users/runner/work/CalibrateEmulateSample.jl/CalibrateEmulateSample.jl/test/MarkovChainMonteCarlo/runtests.jl:12
2022-02-15T16:43:34.9568390Z   Got exception outside of a @test
2022-02-15T16:43:34.9700680Z   AssertionError: isfinite(phi_c) && isfinite(dphi_c)
2022-02-15T16:43:34.9701400Z   Stacktrace:
2022-02-15T16:43:34.9706770Z     [1] secant2!(ϕdϕ::LineSearches.var"#ϕdϕ#6"{Optim.ManifoldObjective{NLSolversBase.OnceDifferentiable{Float64, Vector{Float64}, Vector{Float64}}}, Vector{Float64}, Vector{Float64}, Vector{Float64}}, alphas::Vector{Float64}, values::Vector{Float64}, slopes::Vector{Float64}, ia::Int64, ib::Int64, phi_lim::Float64, delta::Float64, sigma::Float64, display::Int64)
2022-02-15T16:43:34.9707500Z       @ LineSearches ~/.julia/packages/LineSearches/Ki4c5/src/hagerzhang.jl:369
[...]
2022-02-15T16:43:34.9725630Z     [9] optimize_hyperparameters!(gp::GaussianProcess{GPJL})
2022-02-15T16:43:34.9727250Z       @ CalibrateEmulateSample.Emulators ~/work/CalibrateEmulateSample.jl/CalibrateEmulateSample.jl/src/GaussianProcess.jl:184
```
I've definitely hit this "`ϕdϕ`" error before, but I can't remember its cause -- @odunbar, have you come across it in the past? I don't think it's an unlucky random seed in generating the data, as GP regression should be robust to that, but some object is marginally ill-conditioned. 

I can't reproduce the error locally, and [manually re-running](https://github.com/CliMA/CalibrateEmulateSample.jl/actions/runs/1848194576) the suite caused it to pass -- again, all three runs of the suite should have be against identical code from the same commit. Diffing the logs showed a difference in runner (`GitHub Actions 2 (hosted)` vs `Hosted Agent (hosted)`) and image (`macos-11 20220212.1` vs `macos-11 20220207.1`), but I'd hesitate to blame that.

PR #124, opened 2021-12-07
[WIP] Add preconditioned Crank-Nicholson method for MCMC sampling

tsj5:
This PR adds the preconditioned Crank-Nicholson algorithm (pCN) to MarkovChainMonteCarlo.jl. It's selected by passing `alg_type = "pCN"` to `MCMC`. Other aspects of the MCMC interface are unchanged.

Docstrings are updated to reflect this new option. The existing unit test for MCMC estimation is duplicated, and is run for `rwm` as well as `pCN`.

Vanilla random-walk Metropolis sampling (`alg_type = "rwm"`) doesn't have a well-defined limit for high-dimensional parameter spaces; pCN replaces the random walk with an Ornstein–Uhlenbeck [AR(1)] process so that the Metropolis acceptance probability remains non-zero in this limit. See  [Beskos et. al. (2008)](https://www.worldscientific.com/doi/abs/10.1142/S0219493708002378) and [Cotter et. al. (2013)](https://projecteuclid.org/journals/statistical-science/volume-28/issue-3/MCMC-Methods-for-Functions--Modifying-Old-Algorithms-to-Make/10.1214/13-STS421.full).

Useful feedback on this PR was provided by conversations with @odunbar.

### DIFF
--- a/src/MarkovChainMonteCarlo.jl
+++ b/src/MarkovChainMonteCarlo.jl
@@ -50,7 +50,7 @@ struct MCMC{FT<:AbstractFloat, IT<:Int}
     iter::Array{IT}
     "number of accepted proposals"
     accept::Array{IT}
-    "MCMC algorithm to use - currently implemented: 'rmw' (random walk Metropolis)"
+    "MCMC algorithm to use - currently implemented: 'rwm' (random walk Metropolis), 'pCN' (preconditioned Crank-Nicholson)"
     algtype::String
     "Random number generator object (algorithm + seed) used for sampling and noise, for reproducibility."
     rng::Random.AbstractRNG
@@ -85,9 +85,9 @@ function MCMC(
     rng::Random.AbstractRNG = Random.GLOBAL_RNG,
 ) where {FT<:AbstractFloat, IT<:Int}
 
-    
+
     param_init_copy = deepcopy(param_init)
-    
+
     # Standardize MCMC input?
     println(obs_sample)
     println(obs_noise_cov)
@@ -99,7 +99,7 @@ function MCMC(
     println(obs_sample)
     println(obs_noise_cov)
 
-    # We need to transform obs_sample into the correct space 
+    # We need to transform obs_sample into the correct space
     if svdflag
         println("Applying SVD to decorrelating outputs, if not required set svdflag=false")
         obs_sample, unused = Emulators.svd_transform(obs_sample, obs_noise_cov; truncate_svd=truncate_svd)
@@ -107,7 +107,7 @@ function MCMC(
         println("Assuming independent outputs.")
     end
     println(obs_sample)
-    
+
     # first row is param_init
     posterior = zeros(length(param_init_copy),max_iter + 1)
     posterior[:, 1] = param_init_copy
@@ -115,8 +115,10 @@ function MCMC(
     log_posterior = [nothing]
     iter = [1]
     accept = [0]
-    if algtype != "rwm"
-        error("only random walk metropolis 'rwm' is implemented so far")
+    if !(algtype in ("rwm", "pCN"))
+        error("Unrecognized method: ", algtype,
+              "Currently implemented methods: 'rwm' = random walk metropolis, ",
+              "'pCN' = preconditioned Crank-Nicholson")
     end
     MCMC{FT,IT}(obs_sample,
                    obs_noise_cov,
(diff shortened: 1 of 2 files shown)


tsjackson-noaa:
The [test failure](https://github.com/CliMA/CalibrateEmulateSample.jl/actions/runs/1551172081) for 112c06b happened in the [step size test](https://github.com/tsj5/CalibrateEmulateSample.jl/blob/583a40fe22148858873c01e6f47fc119515916db/test/MarkovChainMonteCarlo/runtests.jl#L171). Verified it failed on 1.6 but passes on 1.7. This indicates it's due to the global RNG change in 1.7 (acceptance ratio was close to one of the thresholds used in the logic). The fix in 583a40f leaves logic unchanged but selects a new RNG value for which tests pass on both 1.6 and 1.7.

src/MarkovChainMonteCarlo.jl
```diff
@@ -49,7 +49,7 @@ struct MCMC{FT<:AbstractFloat, IT<:Int}
     iter::Array{IT}
     "number of accepted proposals"
     accept::Array{IT}
-    "MCMC algorithm to use - currently implemented: 'rmw' (random walk Metropolis)"
+    "MCMC algorithm to use - currently implemented: 'rmw' (random walk Metropolis), 'pCN' (preconditioned Crank-Nicholson)"
```
odunbar, line 52: should be 'rwm' my typo I expect.

src/MarkovChainMonteCarlo.jl
```diff
@@ -214,14 +221,20 @@ end
...
+
     if mcmc.algtype == "rwm"
-        prop_dist = MvNormal(zeros(length(mcmc.param)), 
-                             (mcmc.step[1]^2) * proposal_covariance)
+        sample = mcmc.posterior[:,1 + mcmc.iter[1]] .+ rand(prop_dist)
+    elseif mcmc.algtype == "pCN"
+        # 0 < ρ < 1 technically a free parameter; tie to stepsize following
+        # Beskos et al (2017) "Geometric MCMC for infinite-dimensional inverse problems."
+        ρ = (1 - mcmc.step[1]/4) / (1 + mcmc.step[1]/4)
+        sample = ρ * mcmc.posterior[:,1 + mcmc.iter[1]] .+ sqrt(1 - ρ^2) * rand(prop_dist)
```
odunbar, line 234: This is fine. Though you reference Beskos here, I think all that is required here is a map from `f:[0,infty) -> [0,1)` where `f(mcmc.step) = \rho`

src/MarkovChainMonteCarlo.jl
```diff
@@ -214,14 +221,20 @@ end
 
 
 function proposal(mcmc::MCMC)
-
     proposal_covariance = get_cov(mcmc.prior)
- 
+    prop_dist = MvNormal(zeros(length(mcmc.param)),
+                         (mcmc.step[1]^2) * proposal_covariance)
```
odunbar, line 226: I think the the `prop_dist` should be just 
```julia
prop_dist = MvNormal(zeros(length(mcmc.param)), proposal_covariance)
```
then for rwm we have
```julia
sample = mcmc.posterior[:,1 + mcmc.iter[1]] .+ (mcmc.step[1])*rand(prop_dist)
```
otherwise the pCN covariance (which also calls `rand(prop_dist)`) will be incorrectly scaled?

odunbar (commented):
Hi Tom, thanks for this! 

Could you look at my comment on the mcmc.step scaling?

src/MarkovChainMonteCarlo.jl
tsj5: Fixed in d8183f8.

src/MarkovChainMonteCarlo.jl
tsj5: Wording changed in d8183f8 : agree that your condition is necessary, but (unless I was misunderstanding) it's additionally constrained by the (time) discretization scheme used for the Langevin SDE; i.e. if something more complicated than Euler was used, the relation of that stepsize to the SDE's ρ would also need to be more complex. 

At any rate, the intention of the comment was to give the reader a link to follow if they want to know where the equation came from; changed the wording to make this clearer.

src/MarkovChainMonteCarlo.jl
tsj5: Thanks for catching this! Fixed in d8183f8. Was very shortsighted to mindlessly copy the pre-existing code, since stepsize scaling is the **entire point** of implementing pCN!

I don't believe there's a way we could write a unit test to verify this in the future, right? The best we could do would be "run pCN with a small stepsize, and check it doesn't take absurdly long to converge."

tsj5:
bors try

src/MarkovChainMonteCarlo.jl
odunbar: Yeah I don't think there is a way round this over analyzing the mixing times at different dimensions.

src/MarkovChainMonteCarlo.jl
odunbar: I think the beskos one is indeed constrained - but in the MALA case you have the interpretation of the Langevin SDE - here I do not think there is such a clear interpretation?

tsj5:
bors try

odunbar (approved):
LGTM!

PR #132, opened 2022-01-28
[WIP] Substitute types with abstractions

tsj5:
This PR implements the abstract typing done in CliMA/EnsembleKalmanProcesses.jl#100, e.g. `Array{FT, 2}` → `AbstractMatrix{FT}`, in order to be consistent with that dependency.

See the discussion concerning performance at that PR; use of abstract types is [recommended against](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-fields-with-abstract-containers) for perf reasons, but the rationale here is that the code is essentially "glue" rather than numerical routines appearing in hot loops, so writing for generality over perf is justified. 
Another downside is that existing abstract types aren't "abstract" enough, e.g. `UniformScaling` is not a subtype of `AbstractMatrix` and must be handled separately.

As a benefit, the changes made here result in some method signatures being more strongly typed than they are in `master`, allowing us to replace repeated code with multiple dispatch ("Don't Repeat Yourself"). 

`MCMC` is changed to a mutable struct, instead of continuing the current code's practice of enabling mutability by making fields of the struct 1x1 Arrays instead of scalars. This change is a moot point, however, since it will be overridden by PR #130.

### DIFF
--- a/src/Emulator.jl
+++ b/src/Emulator.jl
@@ -7,29 +7,9 @@ using LinearAlgebra
 using DocStringExtensions
 
 export Emulator
-export Decomposition
 
 export optimize_hyperparameters!
 export predict
-
-# SVD decomposition structure
-"""
-    Decomposition{<:AbstractFloat, <:Int}
-
-Struct of SVD decomposition, containing (V,S,Vt), and the size of S, N. 
-"""
-struct Decomposition{FT<:AbstractFloat, IT<:Int}
-    V::Array{FT,2}
-    Vt::Array{FT,2}
-    S::Array{FT}
-    N::IT
-end
-
-# SVD decomposition constructor
-function Decomposition(svd::SVD)
-	# svd.V is of type adjoint, transformed to Array with [:,:]
-	return Decomposition(svd.V[:,:], svd.Vt, svd.S, size(svd.S)[1])
-end
 
 """
     MachineLearningTool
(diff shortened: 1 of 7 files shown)


tsj5:
TODO: add tests with different covariance matrix types

src/Emulator.jl
```diff
@@ -385,22 +352,23 @@ elements on the main diagonal (i.e., the variances), we return the full
...
-    sqrt_singular_values= Diagonal(sqrt.(decomposition.S))
+    sqrt_singular_values = Diagonal(sqrt.(decomposition.S))
     transformed_μ = decomposition.V * sqrt_singular_values * μ
     
     transformed_σ2 = [zeros(output_dim, output_dim) for i in 1:N_predicted_points]
     # Back transformation
     
     for j in 1:N_predicted_points
-        σ2_j = decomposition.V * sqrt_singular_values * Diagonal(σ2[:,j]) * sqrt_singular_values * decomposition.Vt
+        σ2_j = decomposition.U * sqrt_singular_values * Diagonal(σ2[:,j]) * sqrt_singular_values * decomposition.Vt
```
odunbar, line 371: I think this should technically be V not U, as we are doing a coord transformation `(V*S^{1/2})*X*(V*S^{1/2})^T`,

src/GaussianProcess.jl
```diff
@@ -51,13 +53,13 @@ $(DocStringExtensions.FIELDS)
 """
 struct GaussianProcess{GPPackage} <: MachineLearningTool
     "the Gaussian Process (GP) Regression model(s) that are fitted to the given input-data pairs"
-    models::Vector
+    models::Vector{Union{<:GPE, <:PyObject, Nothing}}
```
odunbar, line 56: I think we need to make it more clear in this file where `GPE` comes from - I think it is within `GaussianProcesses.jl`? - so perhaps could we add some description/comment  here about it

odunbar (approved):
Hi Tom - This looks great! I only had a couple of very small comments. 

Beside these I should point out that:
1. I definitely like the removal of decomposition wrapper
2. I like the VecOrMat types to deal with reshaping issues
3. I like the mutability of MCMC (finally we can remove the `[1]` array changes)

Just a small point in the SVD

src/Emulator.jl
tsj5: Thanks for catching this! It was a misunderstand on my part. Fixed in [e9e30bb](https://github.com/CliMA/CalibrateEmulateSample.jl/pull/132/commits/e9e30bb020863835aac117617adbd99435d0a2ad).

src/GaussianProcess.jl
tsj5: Replaced all occurrences of GPE with the fully qualified form GaussianProcesses.GPE in 
b47ba89, hopefully this should make the code self-explanatory.

odunbar:
LGTM!

tsj5:
Update: [5d9def7](https://github.com/CliMA/CalibrateEmulateSample.jl/pull/132/commits/5d9def724ba9795e8a6d2cff9ccda6057943dfe7) is a rebase & squash of commits made up to @odunbar's approval. Extra commits should be non-controversial: 
- `MCMC` struct was made immutable by mistake in a merge; as described above, we make it mutable in this PR with the expectation that it will be replaced by a more robust implementation in #130.
- We support covariance matrices of type `Matrix` (dense), `Diagonal` and `UniformScaling`, for parity with https://github.com/CliMA/EnsembleKalmanProcesses.jl/pull/100. `UniformScaling` is cast to `Diagonal` in the constructor for `Emulator`, since otherwise needed functionality isn't implemented. Identical unit tests are done for these three cases.

tsj5:
bors try

tsj5:
bors try

PR #143, opened 2022-05-28
GP bug for noise_learn = false underprediction of uncertainty

odunbar:
# Purpose
Fix the `noise_learn = false` bug greatly underpredicting uncertainty. In Lorenz example, the following two should be similar. with `noise_learn=false` and `true`:

co-author with @lm2612

# In this PR
- Initial bugfix to correct the regularization noise vs white kernel
- Tighten runtests thanks to improved prediction

### DIFF
--- a/src/GaussianProcess.jl
+++ b/src/GaussianProcess.jl
@@ -192,6 +192,14 @@ function _predict(
     for i in 1:M
         μ[i, :], σ2[i, :] = predict_method(gp.models[i], new_inputs)
     end
+
+    if !(gp.noise_learn)
+        σ2[:,:] = σ2[:,:] .+ 1
+    else
+        magic_number = 1e-3
+        σ2[:,:] = σ2[:,:] .+ magic_number
+    end
+    
     return μ, σ2
 end
 
--- a/test/GaussianProcess/runtests.jl
+++ b/test/GaussianProcess/runtests.jl
@@ -177,10 +177,10 @@ using CalibrateEmulateSample.DataContainers
 
     μ4, σ4² = Emulators.predict(em4, new_inputs, transform_to_real = true)
 
-    @test μ4[:, 1] ≈ [1.0, -1.0] atol = 0.3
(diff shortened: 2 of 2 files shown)


odunbar:
bors try

lm2612:
Looks good with SKLJL. 
When I test with GPJL and `noise_learn=false`, I'm finding the predicted covariance matrix `y_var` to now be around 2x  `Γy`.  Is it possible that this was already included for GPJL?

odunbar:
The predicted covariance (when returned with `transform_to_real=true`) will be the sum of 2 contributions
1. one from the GP kernel approximation (e.g. the fact that we use an RBF to approximate something)
2. one from the observational noise.

So in theory the `y_var` must be >= observational noise. The amount that it is bigger will be due to the error in the GP approximation.

does it return something similar with `noise_learn = true`? This will fix contribution 2. = observational noise, and only learns contribution 1.


Perhaps it would be easier

odunbar:
Ignore what i said: from the GaussianProcess.jl source code I found this: 
thus they add the "lognoise" (i.e. the regularization noise) back in during prediction
```julia
function predict_y(gp::GPE, x::AbstractMatrix; full_cov::Bool=false)
    μ, σ2 = predict_f(gp, x; full_cov=full_cov)
    if full_cov
        npred = size(x, 2)
        return μ, σ2 + ScalMat(npred, exp(2*gp.logNoise))
    else
        return μ, σ2 .+ exp(2*gp.logNoise)
    end
end
```
Thanks! I will update accordingly

lm2612:
Ok great! I forgot to say in my previous comment I was referring to the Lorenz example

odunbar:
OK 
- I've moved the `magic_number` to be an input argument for GP. 
- I've tightened up the new unit tests, so now predicted means, and variances, with noise_learn and without have to be within 5% of each other (was 50% before)

lm2612 (approved):
Great, I get virtually the same results for both GPJL and SKLJL now, thanks!

PR #199, opened 2023-01-12
Modify how to load prior distributions from edmf data

szy21:
## Purpose 

This PR modifies how prior distributions are loaded to be consistent with recent changes in CalibrateEDMF. It also updates the EDMF data in `ent-det-calibration.zip` and `ent-det-tked-tkee-stab-calibration`. 

## Example output
`exp_name = ent-det-calibration`
image
`exp_name = ent-det-tked-tkee-stab-calibration`
image

----


odunbar (commented):
1. Can we remove the manifest completely?  I don't like having them hanging around. (Or is there a reason we had one in this example? I can't recall)
2. Is the difference in the EDMF data just the way it's saved? Otherwise for reference could you post the resulting plot you get from the new data on your PR message. Thanks!

szy21:
> 1. Can we remove the manifest completely?  I don't like having them hanging around. (Or is there a reason we had one in this example? I can't recall)
> 2. Is the difference in the EDMF data just the way it's saved? Otherwise for reference could you post the resulting plot you get from the new data on your PR message. Thanks!

I don't know why we had Manifest for this example, I have deleted it. The EDMF data has also changed (as I don't know exactly how Ignacio generated the data), and I have updated the commit message with the plot.

charleskawczynski:
> (Or is there a reason we had one in this example? I can't recall)

Manifest files are useful for reproducibility

szy21:
> > (Or is there a reason we had one in this example? I can't recall)
> 
> Manifest files are useful for reproducibility

Ah, yes, I did have problems with jld2 compatibility when using a different version of Julia.

odunbar:
Oh definitely, I get the general use for Manifests.

But we have had many more compatatbility issues trying to maintain them, when we had them fixed on GH, than reproducibility issues since we have removed them.

The JLD2 could also be sorted with a compat? Or by saving the data with a newer version of JLD2?

odunbar:
PS @szy21 when you changed the JLD2 stuff, what did you do about the other dataset - the 5 dimensional one? Is it still compatible?

szy21:
@odunbar I updated the 5 parameter calibration data. The results are in the PR message. The parameters do not seem to be well constrained, but that's as expected I think.

odunbar (approved):
@szy21 Thanks! 

As discussed, the example data is not ideal - however, CES seems to be working fine on it. To make plots smoother we can gather more samples:
- Could you change (and ideally rerun) L193 to
`chain = MarkovChainMonteCarlo.sample(mcmc, 200_000; stepsize = new_step, discard_initial = 10_000)`
- Could you double (or measure) the approx run-time in the README.md

After this, merge is good!

odunbar (approved):
LGTM! Thanks

PR #227, opened 2023-07-20
Low rank normalization

### LINKED ISSUE #225: Input normalization fails when ensemble size in EKI less than number of parameters
When normalizing inputs from input–output pairs that come from EKI, if `m` (ensemble size) is less than `p` (number of parameters), then `cov(get_inputs(input_output_pairs), dims = 2)` in [Emulators.jl](https://github.com/CliMA/CalibrateEmulateSample.jl/blob/c8d584810ec5d3de3be8a6b5e2c7c02f94c42312/src/Emulator.jl#L112) is rank-deficient, which causes problems when it is inverted in order to normalize the inputs. Note that even if there are multiple EKI iterations, assuming no localization or additive inflation is applied, then each iteration will still be in the same subspace as the initial ensemble.

As an alternative, the normalization could be performed in the ensemble subspace.

odunbar:
## Purpose 

Closes #225

## Content

- Based on the idea that the covariance defines geometrically a linear transformation from white noise to the data noise, ensemble covariance ``C`` can be viewed as ``C = R S S R^{-1}`` where ``R`` is a rotation, and ``S`` is the sqrt of the singular values of ``C``.  The transformation from white data to the actual data samples is ``C^{1/2} = RS`` and it's "whitening" inverse is ``C^{-1/2} = S^{-1}R^{-1} = S^{-1}R^T`` as ``R`` is a rotation. For rank deficient ``C``, we take the normalization to be instead ``Sinv R^T`` where ``Sinv`` is Diagonal with the first ``rank(C)`` entries equal to ``S^{-1}``, and zero otherwise. 
- Added tests to check the covariances become close to the identity in full or low-dim subspace after normalization
- Changed an input-dim consistency check to account for new dimension change

----

### DIFF
--- a/src/Emulator.jl
+++ b/src/Emulator.jl
@@ -11,6 +11,7 @@ using Random
 
 export Emulator
 
+export calculate_normalization
 export build_models!
 export optimize_hyperparameters!
 export predict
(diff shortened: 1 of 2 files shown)


eviatarbach (approved):
Thank you for addressing this! LGTM.

PR #242, opened 2023-10-26
Joss paper 

### LINKED ISSUE #245: Add the necessary compile tools / markdowns for the JOSS paper in a new branch

### LINKED ISSUE #261: finish JOSS paper Markdown file

odunbar:
## Purpose 

Closes #245
Closes #261

## Content

To Find generated PDF
- go here: [.github/workflows/draft-pdf.yml](https://github.com/CliMA/CalibrateEmulateSample.jl/actions/workflows/draft-pdf.yml)
- Click on latest commit
- Paper is the downloadable "Artifact"

----


lm2612:
on the ABC part
L65 "these tools use statistical emulators to accelerate sampling from a posterior. ABC learns the likelihood directly from the data" 
I could be wrong here but my understanding of ABC was that it doesn't necessarily require an emulator, although looking at gpABC they have the option to use GP to accelerate it. The phrasing "ABC learns the likelihood directly from the data" also confused me slightly as it is normally described as "likelihood free" 

Something along the lines of:
"these tools both approximately sample from the posterior distribution. In ABC, this approximation comes from bypassing the likelihood that is usually required in sampling methods, such as MCMC. Instead, the objective of ABC is that model output agrees with data based on a scalar-valued objective. In CES, the approximation comes from learning the parameter-to-data map, then following this it calculates an explicit likelihood and uses exact sampling via MCMC. Some ABC algorithms also make use of statistical emulators to further accelerate sampling (gpABC)."

odunbar:
This is a fair comment! I have adjusted the paragraph.

odunbar:
NB there are Julia-wide registry issues that are causing test/failures currently. actions can be re-run but buildkite cannot so is being bypassed for now

PR #248, opened 2023-11-02
Sinusoidal example

### LINKED ISSUE #247: Add a simple transparent example to aid the user understanding

lm2612:
## Purpose 
This will close #247

## To-do

- [x] Copy sinusoidal example from EnsembleKalmanProcesses.jl for the calibration step
- [x] Create two emulators (Gaussian process, scalar random feature) and compare performance
- [x] Demonstrate sampling code and plot samples to show how to do uncertainty quantification
- [x] Documentation

## Content

----


docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,415 @@
...
+We ran Ensemble Kalman Inversion with an ensemble size of 10 for 5 
+iterations. This generated a total of 50 input output pairs from our model.
+We will use these samples to train an emulator. The EKI samples make a suitable 
+dataset for training an emulator because in the first iteration, the ensemble parameters 
+are spread out according to the prior, meaning they cover the full support of the
+parameter space. This is important for building an emulator that can be evaluated anywhere 
+in this space. In later iterations, the ensemble parameters are focused around the truth. 
+This means the emulator that will be more accurate around this region.
+
+Load additional packages we need for this section.
```
odunbar, line 217: We load....

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,415 @@
...
+emulator_gp = Emulator(gauss_proc, input_output_pairs, normalize_inputs = true,  obs_noise_cov = Γ)
+optimize_hyperparameters!(emulator_gp)
+```
+For this simple example, we already know the observational noise `Γ=0.2*I`, so we set `noise_learn = false`. 
+However, for more complicated problems we may want to learn the noise as an additional hyperparameter.
+
+We will check performance of the GP by testing on unseen data. We do this by generating some
+additional data from our true model for a set of 10 parameters.
+```
+amplitude_points = range(0.4, 4, length = 10)
```
odunbar, line 250: (Note this is quite a wide range!)

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,415 @@
...
+println("RMSE GP:", rmse_gp)     # [1.0188362036675438; 0.10148451670974754;;]
+```
+
+One metric for assessing emulator performance is to calculate the root mean squared error 
+of our test set, in this case (1.02, 0.10). These errors are relatively small compared against 
+the outputs, around 10%. You could also consider other metrics, such as R^2, mean absolute error or 
+log likelihood. It is also helpful to check emulator performance by plotting the true output 
+against the predicted output. 
+
+![GP_emulator](../assets/sinusoid_GP_emulator.png)
```
odunbar, line 275: I must admit, I find these marginal graphs hard to parse, as you don't know which point on the 1st graph corresponds to which point on the second graph, or in other words, where each evaluation is with respect to the training data. 

Perhaps you could evaluate the GP on a full grid, and plot a heat map of RMSE or something? Similar to the `examples/Regression_2d_2d` in the latest docs?

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,415 @@
...
+
+For a perfect GP emulator, our predictions should line up with the true output, along the y=x dashed line.
+We find that this is mostly the case, except for large values of y_1. This is not too surprising, because
+this region is more extrapolation rather than interpolation, as EKI focuses our training data towards the truth
```
odunbar, line 279: lets' not use words like extrapolation, and just say "points are further from the training set data"

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,415 @@
...
+
+```
+# Extract posterior samples
+posterior = MarkovChainMonteCarlo.get_posterior(mcmc, chain)
+# Back to constrained coordinates
+constrained_posterior = Emulators.transform_unconstrained_to_constrained(
+    prior, MarkovChainMonteCarlo.get_distribution(posterior)
+)
+```
+This gives:
```
odunbar, line 378: This table is not rendered

odunbar (changes requested):
Hi @lm2612 This looks really good so far. I have focussed just on the Docs for now, I'll take a peak at the RF setup etc. in a bit. It's mostly minor comments, although there is one larger comment. 
- I find the marginal GP/RF validation plots quite opaque. It's very hard to interpret how far the data is from the training points etc. as only ordering in one axis is not very clear.

examples/Sinusoid/emulate.jl
```diff
@@ -0,0 +1,241 @@
...
+    "rng" => rng,
+)
+random_features = ScalarRandomFeatureInterface(
+    n_features,
+    input_dim,
+    rng = rng,
+    kernel_structure = kernel_structure,
+    optimizer_options = optimizer_options,
+)
+emulator_random_features = Emulator(random_features, input_output_pairs, normalize_inputs = true)
```
odunbar, line 170: we provide `obs_noise_cov = Γ` to GP but not RF - therefore it does not rescale with the SVD properly. This is not a fair experiment.

examples/Sinusoid/emulate.jl
```diff
@@ -0,0 +1,241 @@
...
+optimize_hyperparameters!(emulator_random_features)
+
+# Test on new data
+y_rf, y_rf_covs = Emulators.predict(emulator_random_features, x_test_unconstrained);
```
odunbar, line 174: adding the obs_noise_cov means that we now transform the data. we need to add `transform_to_real=true` (as we do with GP) in the predict methods when we calculate RMSE or plots

examples/Sinusoid/emulate.jl
```diff
@@ -0,0 +1,241 @@
...
+savefig(p, joinpath(data_save_directory, "sinusoid_GP_vs_RF_emulator.png"))
+
+# This comparison shows the GP emulator is more accurate on the test data compared to the RF emulator.
+# Both methods are more accurate around y_obs=(6.3, 7.1) with worse performance at the edges.
+# This is as expected, because our training dataset from EKI has more points focused around theta_true.
+
+# Plot outputs as contours
+# f returns emulator prediction only (no variance) given x1 and x2, for output j=1 or 2
+f(x1, x2; j = 1) =
+    Emulators.predict(emulator_random_features, Emulators.transform_constrained_to_unconstrained(prior, hcat(x1, x2)'))[1][
```
odunbar, line 206: adding the obs_noise_cov means that we now transform the data. we need to add `transform_to_real=true` (as we do with GP) in the predict methods when we calculate RMSE or plots

examples/Sinusoid/sample.jl
```diff
@@ -0,0 +1,214 @@
...
+obs_file = joinpath(data_save_directory, "observations.jld2")
+y_obs = load(obs_file)["y_obs"]
+theta_true = load(obs_file)["theta_true"]
+Γ = load(obs_file)["Γ"]
+# Get emulators and prior
+emulator_file = joinpath(data_save_directory, "emulators.jld2")
+emulator_gp = load(emulator_file)["emulator_gp"]
+prior = load(emulator_file)["prior"]
+
+init_sample = vec([1.0, 1.0])
```
odunbar, line 39: It's good practice to start the markov chain where the emulators are accurate - e.g. among the training points ( e.g. the unconstrained mean of the final ensemble is a good place). This will reduce spin-ups and help with step-size optimization

examples/Sinusoid/emulate.jl
```diff
@@ -0,0 +1,241 @@
...
+
+## Random Features
+# An alternative to Gaussian process for the emulator
+# We have two input dimensions and two output dimensions.
+input_dim = 2
+output_dim = 2
+# Select number of features
+n_features = 60
+nugget = 1e-12
+kernel_structure = SeparableKernel(LowRankFactor(2, nugget), OneDimFactor())
```
odunbar, line 154: A quick play found that `LowRankFactor(1,nugget)` is enough here (rather than 2)

examples/Sinusoid/emulate.jl
```diff
@@ -0,0 +1,241 @@
...
+
+# Create random features
+optimizer_options = Dict(
+    "n_ensemble" => 20,
+    "scheduler" => EKP.DataMisfitController(on_terminate = "continue"),
+    "n_iteration" => 50,
```
odunbar, line 160: add "verbose" => true, this will print errors so you can easily see what a reasonable number of iterations/ensemble size is

odunbar (changes requested):
Hi @lm2612 

A review on the example itself: 
- The difference in performance is due to it not being a fair experiment, I have detailed the small fix, and I observe this balances out the behaviour significantly. though the RF still has a heavier tail for theta_1 (Note RF still learns the noise where GP doesnt)

odunbar:
Another Kernel that works even better is the following vector RF case without decorrelation: 
```
kernel_structure = NonseparableKernel(LowRankFactor(2,nugget))
optimizer_options = Dict(
    "scheduler" => EKP.DataMisfitController(on_terminate = "continue"),
   "n_iteration" => 50,
    "rng" => rng,
    "verbose"=>true,
)
random_features = VectorRandomFeatureInterface(
    n_features,
    input_dim,
    output_dim,
    rng = rng,
    kernel_structure = kernel_structure,
    optimizer_options = optimizer_options,
)
emulator_random_features = Emulator(random_features, input_output_pairs, normalize_inputs = true, obs_noise_cov=Γ, decorrelate=false)
```

examples/Sinusoid/emulate.jl
lm2612: Good point, thanks for spotting this, we should definitely use the same for both!

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,425 @@
...
+
+
+| parameters         | mean    | std |
+| :---------------- | :------: | :----: |
+| amplitude         |   3.0324 | 0.2863  |
+| vert_shift        |   6.3753 | 0.4591  |
+
+This is in agreement with the true $\theta=(3.0, 7.0)$ and with the observational covariance matrix we provided $\Gamma=0.2 * I$ (i.e., a standard deviation of approx. $0.45$). `CalibrateEmulateSample.jl` has built-in plotting
+recipes to help us visualize the prior and posterior distributions.
```
odunbar, line 398: Highlight that it plots the marginal distributions only

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,425 @@
...
+# We can quickly plot priors and posterior using built-in capabilities
+p = plot(prior)
+plot!(posterior)
+
+```
+![GP_posterior](../assets/sinusoid_posterior_GP.png)
+
+
+The MCMC has learned the posterior distribution which is much narrower than the prior. 
+For multi-dimensional problems, we may be interested in how the posterior distributions for
```
odunbar, line 410: For multidimensional problems, we may be interested in how ... 

Replace with

For multidimensional problems, the posterior is typically multidimensional, and marginal distribution plots do not show how parameters co-vary, we plot...

docs/src/examples/sinusoid_example.md
```diff
@@ -0,0 +1,425 @@
...
+
+### Sample with Random Features
+
+We can repeat the sampling method using the random features emulator instead of the Gaussian
+process and we find similar results: 
+
+![RF_2d_posterior](../assets/sinusoid_MCMC_hist_RF.png)
+
+It is reassuring to see that our uncertainty quantification methods are robust to the different emulator
+choices here. The GP and RF emulators showed similar accuracy during validation, and therefore the MCMC using
```
odunbar, line 424: I think we should be honest here, and state that one can obtain poor results if the GP kernel or RF kernel structure is chosen to be too inflexible. So people should try several configurations out too

docs/make.jl
```diff
@@ -8,6 +8,7 @@ using Documenter
 examples = [
     "Lorenz example" => "examples/lorenz_example.md",
     "Turbulence example" => "examples/edmf_example.md",
+    "Sinusoid example" => "examples/sinusoid_example.md",
```
odunbar, line 11: Can we call this "Simple example walkthrough" and move it to the top of the `examples[...]` list (so it is the first examples page)

odunbar (approved):
LGTM - this is a fantastic example! Please run the julia formatter - otherwise my comments are very minor.

Then we can squash&merge

PR #253, opened 2023-11-13
[WIP] Resuscitate the Cloudy example

### LINKED ISSUE #256: Write consistent docs page for Cloudy

bielim:
## Purpose 
Closes #256 

Resuscitate the Cloudy example. The example demonstrates the full CES pipeline using a simple microphysics model but is currently deprecated.
The goal is to get it working again, and to include a version with a Random Feature emulator in addition to the currently implemented Gaussian Process emulator.
Documentation will be added as well. 

Docs can be found [here](https://clima.github.io/CalibrateEmulateSample.jl/previews/PR253/examples/Cloudy_example/)

## Content
- [x] Get the deprecated example to work again
- [x] Add a version using a RF emulator
- [x] Clean-up 
- [x] Documentation


docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+include(joinpath(@__DIR__, "DynamicalModel.jl"))
+```
+
+and finally the EKP packages.
+
+```julia
+using EnsembleKalmanProcesses
+using EnsembleKalmanProcesses.Observations
+using EnsembleKalmanProcesses.ParameterDistributions
+using EnsembleKalmanProcesses.DataContainers
```
odunbar, line 93: this isn't in the project toml, should it be?

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+We use `constrained_gaussian` to add the desired scale and bounds to the prior distribution, in particular we place lower bounds to preserve positivity (and numerical stability). 
+
+```julia
+# We choose to use normal distributions to represent the prior distributions of
+# the parameters in the transformed (unconstrained) space.
+prior_N0 = constrained_gaussian(param_names[1], 400, 300, 0.4 * N0_true, Inf)
+prior_θ = constrained_gaussian(param_names[2], 1.0, 5.0, 1e-1, Inf)
+prior_k = constrained_gaussian(param_names[3], 0.2, 1.0, 1e-4, Inf)
+priors = combine_distributions([prior_N0, prior_θ, prior_k])
+```
```
odunbar, line 139: you could put in the command `plot(prior)` and the figure (given in physical space by default), here.

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+The optimal parameter vector determined by the ensemble Kalman inversion is the
+ensemble mean of the particles after the last iteration, which is printed to
+standard output. An output directory is created, where a file
+`cloudy_calibrate_results.jld2` is stored, which contains all parameters and
+model output from the ensemble Kalman iterations (both as
+`DataContainers.DataContainer` objects), the mean and one sample of the
+synthetic observations, as well as the true prameters and their priors. In
+addition, an animation is produced that shows the evolution of the ensemble of
+particles over subsequent iterations of the optimization, both in the
+computational (unconstrained) and physical (constrained) spaces.
```
odunbar, line 225: You can put the visualization here if you like? Would be the right place for it. you just include the `.gif` as you would do a figure

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+# Wrapper for GP
+gaussian_process = GaussianProcess(
+    gppackage;
+    kernel = gp_kernel,
+    prediction_type = pred_type,
+    noise_learn = false,
+)
+
+# Use all input-output pairs available from the EKI iterations
+input_output_pairs = get_training_points(ekiobj,
```
odunbar, line 332: These are the same for both options. as will the emulator (see my comment on the example) so really the GPJL and SRFI ones should only differ in the creation of the ML tool specifically that gets passed into the emulator

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+new_step = optimize_stepsize(
+    mcmc;
+    init_stepsize = 0.1,
+    N = 2000,
+    discard_initial = 0
+)
+```
+
+We choose a sample size of 100'000 for the actual MCMC, discarding the first
+1'000 samples as burn-in:
```
odunbar, line 443: 100,000 or 1,000 is more american :P

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+    truth_sample,
+    priors,
+    emulator;
+    init_params = u0
+)
+
+new_step = optimize_stepsize(
+    mcmc;
+    init_stepsize = 0.1,
+    N = 2000,
```
odunbar, line 437: maybe note the stepsize is chosen by running chains of length 2000

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+### Results
+The following three plots show the posterior distributions obtained by the CES
+algorithm, for the inverse problem of finding the parameters ``N_{0,
+0}``, ``k_0``, and ``\theta_0``, which define a gamma distribution of droplet
+masses in Cloudy at time ``t = 0``. The results are shown in the unbounded space
+where the CES algorithm takes place. The samples posterior distributions are
+shown as grey histograms, the parameters' prior distribution are shown in blue,
+and the true parameter values are marked as vertical green lines.
+
+![posterior_N0_rf](../assets/posterior_rf-scalar_N0.png)
```
odunbar, line 478: could we plot both the rf-scalar and a GP run here? If GPJL has issues, maybe we can use scikit-learn?

odunbar (commented):
Thanks very much for the hard work! This looks awesome! I have some small comments and a couple of larger points:
- Could you plot corner for the results- i have a snippet you can copy paste for this (rather than marginals just to emphasise that the posterior is a joint distributions)
- I think the Manifest should be removed, and the Project toml can pin a cloudy version, and can add in EKP and/or CES versions (note that EKP is also within CES so you could just import CES and do `using CalibrateEmulateSample.EnsembleKalmanProcesses`) for example
- the modularity for GP & RF should only be in the ML tool, the `Emulator` object should be constructed same for both options (and shouldn't effect the results)

examples/Cloudy/Cloudy_emulate_sample.jl
```diff
@@ -0,0 +1,291 @@
...
+        println(post_cov)
+
+        # Plot the posteriors together with the priors and the true parameter
+        # values (in the transformed/unconstrained space)
+        n_params = length(get_name(posterior))
+
+        gr(size = (800, 600))
+
+        for idx in 1:n_params
+            posterior_samples = dropdims(
```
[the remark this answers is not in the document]
bielim: I added the joint distribution correlations in physical space, and also changed the marginals from unconstrained/computational space to constrained/physical space (the docs are currently showing both types of plots---not sure if you wanted me to remove the marginals altogether, if so please let me know!)
The only plot that is in the unconstrained space is the priors. It just didn't look nice in the constrained space (see attached pic)---making it look nice would probably require going in and tweaking the plot recipe.
cloudy_priors_constrained

docs/src/examples/Cloudy_example.md
```diff
@@ -0,0 +1,518 @@
...
+
+We will show how Calibrate-Emulate-Sample (CES) can be used to learn the parameters of the initial cloud droplet mass distribution from observations of the moments of that mass distribution at a later time. 
+
+Cloudy is used here in a "perfect model" (aka "known truth") setting, which
+means that the "observations" are generated by Cloudy itself, by running it with
+the true parameter values---in more realistic applications, one would use actual
+measurements of cloud properties.
+
+The schematic below gives an overview of the example: The input to the CES
+algorithm consists of data ``y``, the observational covariance ``Γ``, and prior
```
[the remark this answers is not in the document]
bielim: I moved the explanation after the schematic.

docs/src/examples/Cloudy_example.md
bielim: I added a comment explaining this: "Before running the actual MCMC chain, we determine a good step size by running chains of length `N = 2000`."

examples/Cloudy/Cloudy_emulate_sample.jl
```diff
@@ -0,0 +1,291 @@
...
+            srfi = ScalarRandomFeatureInterface(
+                n_features,
+                n_params,
+                kernel_structure = kernel_structure,
+                optimizer_options = optimizer_options,
+            )
+
+            retained_svd_frac = 1.0
+
+            emulator = Emulator(
```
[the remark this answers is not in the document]
bielim: Agreed! I re-structured the example such that the `Emulator` construction is now the same for both RF and GP, with the `mlt` input holding the information on which tool is used. I had set the normalization of inputs to `false` for the GP emulator because for some reason this improves the emulator performance substantially. Do you have an explanation for this? If we want to maximize the quality of the emulator I will have to set normalize to `true` or `false` depending on whether we're dealing with RF or GP. This would be less elegant than having the `mlt` be the only difference between the two in terms of the `Emulator` construction...do you have thoughts on why for the GP case normalizing the inputs leads to worse results than not normalizing them?

docs/src/examples/Cloudy_example.md
bielim: Yep, I restructured the example such that the `input_output_pairs` only get defined once (see comment above)

docs/src/examples/Cloudy_example.md
bielim: Yes we can, I just thought that showing a corner plot and the three marginals for each emulator would be a bit of a plot overload, especially given that readers of the docs will probably be interested more in the functionality / step-by-step guidance than in the results.
In the current version of the docs, I show the corner plots for both GP and RF, and the three marginals for the GP case---let me know if you think we should add more!

examples/Cloudy/Cloudy_emulate_sample.jl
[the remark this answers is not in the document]
odunbar: Good point - I guess some of these are quite heavy tailed..

odunbar (commented):
This looks great now thank you!

- One very small comment, could you label the figures in assets with `cloudy_...` just so we can distinguish them?
- Can you go to your base of your CES repo and call 
`julia --project=.dev .dev/climaformat.jl` to make the formatter happy, 

It would be good to see the vector case in here. I'd advocate for 
`NonSeperable(LowRankFactor(r))` for some low-ish rank `r`
and you can set decorrelate = false in the emulator for this case (so it doesn't use the SVD). However, if this is too much work, then I would also be happy to see this merged after the above small comments.

examples/Cloudy/Project.toml
```diff
@@ -12,7 +12,6 @@ Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
-Sundials = "c3572dad-4567-51f8-b174-8c6c989267f4"
 
 [compat]
 julia = "~1.6"
```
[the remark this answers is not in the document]
bielim: Done!

docs/src/examples/Cloudy_example.md
bielim: Yes, good point, I added that too.

odunbar:
I'm removing the .dev Manifest, and then will merge of success

PR #262, opened 2023-12-11
Fix windows test failures

odunbar:
## Purpose 

Posted this issue to `SciKitLearn.jl` as it appears like it is on their side.

## Content

- Running with newer test pipelines
- Running with newer versions of python packages
- add a more noticeable test for when python package updates fail

----


odunbar:
CI and test-pipeline changes, leading to passing tests with latest python and on all platforms with latest julia versions

PR #265, opened 2023-12-17
Add vector RF case to Cloudy example

bielim:
## Purpose 
Add a vector Random Feature (RF) emulator to the Cloudy example (in addition to the already existing emulator options of a scalar Random Feature emulator and a Gaussian Process emulator)

## To-do

- [x] Update docs to include vector RF case

## Content
- [x] Implement vector RF case

----
- [x] I have read and checked the items on the review checklist.


bielim:
These are the posteriors (marginals and pair plots) resulting from the vector RF case:

pairplot_posterior_constr png_rf-nosvd-nonsep
marginal_posterior_constr_rf-nosvd-nonsep_k
marginal_posterior_constr_rf-nosvd-nonsep_θ
marginal_posterior_constr_rf-nosvd-nonsep_N0

odunbar:
Hi @bielim 
One thing I noticed when rerunning this example is that the Project.toml somehow has 
- two lines called `Cloudy` 
- no FFTW but a compat for it 
Perhaps this got muddled in the merging somewhere - likely my mistake.

bielim:
> Hi @bielim One thing I noticed when rerunning this example is that the Project.toml somehow has
> 
> * two lines called `Cloudy`
> * no FFTW but a compat for it
>   Perhaps this got muddled in the merging somewhere - likely my mistake.

No problem, I'll fix that.

odunbar:
Until Cloudy v0.3.0 becomes a Julia release, note that `add Cloudy#f977fff` works, but Cloudy#main does not work currently

odunbar:
Hi @bielim -  I think there can be an issue with the points used for training data (i expect this will need further work down the line anyway).  

My guess is that for some reason the previous setup did a very poor train/validate split. something we can investigate in the future... However, if you take fewer training points 
```julia
input_output_pairs = get_training_points(ekiobj, length(get_u(ekiobj)) - 2) 
# not - 1 at the end...
``` 
you get something more like the original plots:
pairplot_posterior_constr png_rf-nosvd-nonsep

PS and in this case it seems this problem really doesnt need many features at all: the following used only `n_features=50` for the nosvd-nonsep case
pairplot_posterior_constr png_rf-nosvd-nonsep

bielim:
> Hi @bielim - this seems to be an issue with the points used for training data.
> 
> My guess is that for some reason the previous setup did a very poor train/validate split. something we can investigate in the future... However, if you take fewer training points
> 
> ```julia
> input_output_pairs = get_training_points(ekiobj, length(get_u(ekiobj)) - 2) 
> # not - 1 at the end...
> ```

The crux is to find `input_output_pairs` that work for all three cases (scalar RF, GP, vector RF)---for some reason the GP case seems to work a lot better when using training points from all iterations. Or should I drop the (self-imposed) constraint that all cases have to use the same training points? It seemed simpler/more elegant, but in practice you probably would try to different numbers of iterations/training points here...

bielim:
> Hi @bielim - this seems to be an issue with the points used for training data.
> 
> My guess is that for some reason the previous setup did a very poor train/validate split. something we can investigate in the future... However, if you take fewer training points
> 
> ```julia> ```julia
> input_output_pairs = get_training_points(ekiobj, length(get_u(ekiobj)) - 2) 
> # not - 1 at the end...
> ```
 I changed the number of iterations/training points used to `length(get_u(ekiobj)) - 2` but kept the number of features at 400 (seems to be the better number to use for both RF cases---for some reason the scalar RF didn't work so well with only 50). Plus runtime(RF) <<<<<<< runtime(GP) anyway, whether we use 50, 400 or 1000 features :-D

bielim:
> Cloudy#f977fff

hmmm....I thought that `add Cloudy#main` had worked for me...

odunbar (approved):
Thanks for the work, LGTM! Please resolve the formatter first and change the PR title before the squash&merge!

PR #272, opened 2024-01-03
Add figures to EDMF example

### LINKED ISSUE #273: add figures to EDMF examples, adjust for JLD2 updates

odunbar:
## Purpose 

Close #273

NB this is still probably the weakest example as it doesn't have a walkthrough, and due to the computational expense I expect most users would not try to run it. but it is useful to keep and has some expected results now at least.

## Content

- add images in docs
- reduce example to just a couple of cases  

Docs found [here](https://clima.github.io/CalibrateEmulateSample.jl/previews/PR272/examples/edmf_example/)

----


odunbar:
merging, this just adds premade figures, and removes code-branches that are not run

PR #277, opened 2024-01-24
Bugfix parameter name bug for MV distributions, cap mac-os tests to julia v1.9.0

odunbar:
## Purpose 

MCMCChains requires one name per dimension for MCMC diagnostics. This causes an error when Multivariate distributions are used. This PR fixes this bug.

## Content

- uses the dimension of prior distribution in MCMC to clone the name e.g. a 3D "u" -> ["u_1","u_2","u_3"] 
- adds a quick unit-test for a case where #names < #prior dimensions

## Misc
- using `sqrt(C) * MVNormal(0,I)` instead of `MVNormal(0,C)`
- for an unknown reason, the Mac tests with v1.9.1 onwards are hanging on the package build of CES, so capping at v1.9.0

----
- [x] I have read and checked the items on the review checklist.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -80,8 +80,8 @@ jobs:
     - name: Set up Julia
       uses: julia-actions/setup-julia@v1
       with:
-        version: 1.9.0
-        
+        version: 1.9.1
+                 # later versions causes hanging in package building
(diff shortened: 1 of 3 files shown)


odunbar:
Small bugfixes with unit tests, and buildkite passing with unchanged results.

PR #347, opened 2025-01-24
Compat with EKP 2.1.0

odunbar:
## Content

- remove call to `approximate_corr_std` that no longer exists. Replace with new simple expression found in `EnsembleKalmanProcesses.jl v2.1.0`, file  `src/Localizers.jl`

----


odunbar:
Merging on passed tests, as buildkite is queuing for too long.

PR #349, opened 2025-01-29
WIP: Enable (naive) minibatching within MCMC.

odunbar:
## Content

- New method to decorrelate a vector of data sample vectors
- New emulator posterior function that still enables us to build the (cumbersome) AdvancedMH.DensityModel object. using by defining `x -> f(x, emulator, prior, data)` in the factory
- emulator posterior now taks 1/n * sum(log(p(y_i|x))) + log_prior(x)` over the data vector `y_i in Y`. This is a naive formulation, requiring `n` evaluations of the emulator for every MCMC step.
- added a unit test where 100 slightly perturbed data samples, and acheive the same target.

## Misc
The net result of [(Barbenet, Doucet, Holmes 2017)](https://jmlr.org/papers/v18/15-205.html) )  is that minibatching within MCMC is a very technically complex problem, and for computational benefit without loss of accuracy requires a nested procedure to approximate the accept-reject mechanicsm, and futher surrogates of the log-likelihood. It may be that simple parallelization or alternative (parallel) sampling methods would be preferred
- The `ObservationSeries` object contains both data samples and noise covariance. Here it appears that MCMC requires only the data samples, while the emulator uses the noise covariance (for decorrelation), this should be addressed in a future issue to do with data processing but not here

----

### DIFF
--- a/docs/src/API/MarkovChainMonteCarlo.md
+++ b/docs/src/API/MarkovChainMonteCarlo.md
@@ -8,7 +8,14 @@ CurrentModule = CalibrateEmulateSample.MarkovChainMonteCarlo
 
 ```@docs
 MCMCWrapper
-MCMCWrapper(args...; kwargs...)
+MCMCWrapper(
+    mcmc_alg::MCMCProtocol,
+    observation::AMorAV,
+    prior::ParameterDistribution,
+    em::Emulator;
+    kwargs...,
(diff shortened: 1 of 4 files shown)


odunbar:
passes tests, and buildkite. Docs succeed and  Increases code-cov, therefore will merge without review

PR #360, opened 2025-05-09
[WIP] dimension reduction playground

odunbar:
## Content

----


ArneBouillon:
After the refactor:
- We now have both a linear-exponential and a Lorenz test case
- We can compare the reduced posteriors using either EKS or MCMC
- The settings have been centralised in the file `settings.jl`
- Data and figure files are in their own folders, to keep everything clean
- I changed the `.jl` files to start with `step1`, `step2`, and `step3`, to hopefully make the code easier to approach

ArneBouillon:
I currently changed the projection from/to the reduced spaces to purely using `U_r`/`U_r'`/`V_r`/`V_r'`.

I believe this is equivalent to what we did before, except for setting the coefficients of the "small" eigenvectors to zero instead of setting them to the corresponding (small) eigenvalues. I'm not sure which makes more sense.

odunbar:
Lets close for now, we can merge examples separately as needed

PR #371, opened 2025-07-08
Re-organize data reduction

ArneBouillon:
This PR makes the following changes:
 - The `orig2zscore` and `zscore2orig` functions, which were unused, are removed.
 - The individual data processors are moved to their own files.
 - `Decorrelater` is renamed to `Decorrelator`
 - Instead of exporting all subroutines related to data processing, I now only export `create_encoder_schedule`, `initialize_and_encode_with_schedule!`, `encode_with_schedule`, and `decode_with_schedule`. My rationale for this is twofold.
  - It makes clear that these are the APIs that should be used in practice. The internal functions now start with an underscore (can be removed! just an idea) and can, if necessary, still be used outside the package as, e.g., `CES.Utilities._initialize_and_encode_data!`.
  - It allows the internal subroutines to implement only the methods necessary for our API functions, reducing the number of methods we need to define and maintain.
 - Partially thanks to the above change, I could remove the `extract_data` functions.

I also started working towards `LikelihoodInformed` processors by passing both structure matrices to the initialization code of `PDCP`s. I didn't pass the observation yet, since I'm still looking for a clean way to add this to the pipeline.


odunbar:
I have addressed some of your comments in the PR so there may be small conflicts - I will refrain from further changes until this is merged. (and will work on a Docs page)

odunbar (approved):
LGTM!

PR #376, opened 2025-08-20
Add likelihood-informed data processors

### LINKED ISSUE #370: Add new utilities for the likelihood-informed DataProcessor
## Feature

- the data/likelihood informed processor struct and associated methods
- access to _both_ input and output structure matrices on calling `initialize_processor!` for the `PairedDataContainerProcessor`
- an `output_structure_vector` a.k.a the observation itself, (or perhaps the mean of the data samples as a proxy) should be passable into the emulator and processed as the strcuture matrices do (possibly also do this with the input?)

### LINKED ISSUE #380: Add a machine learning tool for user-defined functions
## Issue

Primarily for debugging etc. It would be convenient to define a new "machine learning tool" type that just evaluates a given function. We will not directly address the dimension reduction handling here, so the user would have to know about this.  But the idea is that one could e.g. define the actual forward map "G" in simple settings and pass this in.

## Possible solution
- define a new tool called "UserDefinedMLT" which is built with a function,  
- optimize does nothing and predict just calls the function and uses the result to produce a mean and covariance

ArneBouillon:
Implement the likelihood-informed data processor from our Overleaf

- Closes #370
- Closes #380

This PR makes the following changes.
- We add a `LikelihoodInformed` data processor, which combines input data, output data and the actual inverse problem to find good reduced spaces.
- We add an (undocumented and inconvenient to access) option to construct a `Decorrelator` that uses a fixed reduced-space dimension, instead of calculating it based on a variance threshold. This helps primarily when testing or doing comparisons.
- We add a machine learning tool that simply calls a user-defined function on the decoded input and encodes the result again (closing #380)
    - This is currently in `examples/DimensionReduction/emulate_sample_linlinexp.jl`. I feel like we need to make it a bit more versatile and less hacky before extracting it from there.
    - @odunbar what do you think?
- We add an example comparing `Decorrelator + LikelihoodInformed` to just `Decorrelator`.
    - TODO: I'm working on more experiments.

Big TODO: The error estimate for output space reduction when `α ≠ 0` is way off, which would result in high errors when using `retain_KL` as a criterion to find a reduced space. I'm still looking into why this is.

There are some test failures that I still have to look at, but I think this PR is ready to review.


examples/DimensionReduction/emulate_sample_linlinexp.jl
```diff
@@ -0,0 +1,142 @@
...
+
+            post_mean = mean(posterior)
+            post_cov = cov(posterior)
+
+            post_means = hcat(post_means, reshape(post_mean, input_dim, 1))
+            push!(post_covs, post_cov)
+        end
+
+        all_errs[dim_i, :] =
+            [norm(post_means[:, 2] - v) / norm(post_means[:, 2]) for v in eachcol(post_means[:, 3:end])]'
```
odunbar: Why is the reference emulator not used here? Otherwise this is just showing the size of difference to a PCA-truncation mean.

PS. Running the example gave:

```5×6 Matrix{Float64}:
 0.771104  0.724079  0.686967  0.675671  0.720269  0.687709
 0.397877  0.567556  0.629621  0.656653  0.666858  0.66542
 0.40432   0.393665  0.350307  0.287113  0.297794  0.238406
 0.436975  0.370324  0.34126   0.278414  0.262382  0.233595
 0.376547  0.295315  0.151808  0.147098  0.094317  0.118197
```

src/Utilities/likelihood_informed.jl
```diff
@@ -0,0 +1,251 @@
...
+$(TYPEDEF)
+
+Uses both input and output data to learn a subspace that allows for a reduced posterior which is close to the full posterior.
+
+Preferred construction is with the [`likelihood_informed`](@ref) method.
+
+# Fields
+$(TYPEDFIELDS)
+"""
+mutable struct LikelihoodInformed{FT <: Real} <: PairedDataContainerProcessor
```
odunbar: In general in the package we have never used mutable structs, only immutable ones. Which is perhaps annoying, but for consistency could we stick with immutable for now and do a larger change later (sorry)?

I also saw a couple of issues like this [ where the union {Nothing,X} do weird things before mutation](https://github.com/JuliaLang/julia/issues/58470). Not a big deal but happens quite commonly

src/Utilities/likelihood_informed.jl
```diff
@@ -0,0 +1,251 @@
...
+        end
+        noise_cov_inv = inv(obs_noise_cov)
+
+        li.apply_to = apply_to
+
+        grads = if li.grad_type == :linreg
+            grad = (samples_out .- mean(samples_out; dims = 2)) / (samples_in .- mean(samples_in; dims = 2))
+            fill(grad, size(samples_in, 2))
+        else
+            @assert li.grad_type == :localsl
```
odunbar: Could we have useful `throw(ArgumentError(...))`s for the user over asserts? In our experience it is much better to tell the user to use r to provide the correct args than to have asserts.

src/Utilities/likelihood_informed.jl
```diff
@@ -0,0 +1,251 @@
...
+                            noise_cov_inv *
+                            ((1 - α)obs_noise_cov + α^2 * (y - g) * (y - g)') *
+                            noise_cov_inv *
+                            grad for (g, grad) in zip(eachcol(samples_out), grads)
+                        ),
+                    ),
+                    sortby = (-),
+                )
+            else
+                @assert apply_to == "out" && α ≈ 0 && obs_whitened
```
odunbar: likewise assert statement above... and others below

examples/DimensionReduction/emulate_sample_linlinexp.jl
```diff
@@ -0,0 +1,142 @@
...
+
+
+rng = Random.MersenneTwister(41)
+input_dim = 100
+output_dim = 100
+αs = [0.0, 0.25, 0.5, 0.75, 1.0]
+
+num_trials = 1
+for trial in 1:num_trials
+    loaded = load("datafiles/ekp_linlinexp_$(trial).jld2")
```
odunbar: I found that loading this in a new session fails, it seems like the only time it worked was if I ran both `calibrate` and `emulate_sample` in the same session.

odunbar (commented):
So far looks great .

I can see where you butted up against missing features/API. Happy to have resolved here:
- The dim criterion for decorrelator -> Happy to resolve here
- The No-Emulator `MachineLearningTool` -> Happy to resolve here
- encode_data/decode_data for `Vector` types -> Happy to resolve here

Maybe can be resolved later:
- Parallel MCMC, , I have created an issue to expand this so you don't have to reach into internals to do parallel MCMC -> We can resolve this in another PR if that's useful
- The EKP scheduler with a checkpoint -> We can create a PR for this in EKP (may avoid the issue below)

Regarding the examples.
- I had to rebase this branch with `main` to get the MCMC working again
- It is a little hard to interpret what the final result of `emulate_sample_linlinexp.jl` does. it also seemed like to run `emulate_sample` you always had to run `calibrate` in the same session (I wonder if this has something to do with the CheckpointScheduler definition)
- I haven't yet run lorenz

odunbar:
Closing due to successor #410 merging. Note there may still be advances here we could use in future.

PR #410, opened 2026-04-01
[Updated] Likelihood-informed processors

### LINKED ISSUE #370: Add new utilities for the likelihood-informed DataProcessor

### LINKED ISSUE #419: Counterintuitive value i of get_training_points(ekp, i)
# Issue

For extraction of itrations from ekp. it would make sense that if you provide $i=3$ to imply extract 3 iterations, that these would be iterations `1:3`, 

However, iterations extracted is in fact `N_it - 3:N_it` The "latest" 3 iterations.

This is not clearly communicated to the user. and should be changed if not at least messaged as it could confuse interpretation.

## Possible solutions 

The following function would make more sense. it takes `1:3` by default, and also will tell the user which iterations they are extracting when they define an integer input

```julia
function get_training_points(
    ekp::EKP.EnsembleKalmanProcess{FT, IT, P},
    train_iterations::Union{IT, AbstractVector{IT}};
    g_final = nothing,
) where {FT, IT, P}

    if !isa(train_iterations, AbstractVector)
        iter_range = 1:min(train_iterations, length(get_u(ekp)))
    else
        min_it = max(1,minimum(train_iterations))
        max_it = min(length(get_u(ekp)),maximum(train_iterations))
        iter_range = [i for i in train_iterations if  (min_it <= i <= max_it) ]
    end
    
    @info "extracting iterations $(iter_range) from EnsembleKalmanProcess"
    if (maximum(iter_range) > length(get_g(ekp))) && isnothing(g_final)
        throw(ArgumentError("Require outputs associated to parameters `get_u(ekp,$(maximum(iter_range)))`. \n the user can provide this with the `g_final` keyword. Received `g_final=$(g_final)`"))
    end
        
    u_tp = []
    g_tp = []

    g_len = length(get_g(ekp))
    g_full = [get_g(ekp, i) for i in iter_range[iter_range .<= g_len]]
    if !isnothing(g_final)
        if (isa(g_final, AbstractMatrix)) # add the matrix
            if !(size(g_full[1]) == size(g_final))
                throw(
                    ArgumentError("Expected `g_final` size: $(size(g_full[1])), received $(size(g_final)).")
                )
            end
            push!(g_full, g_final)
        else
            throw(
                ArgumentError("Expected `g_final` to be type `<:AbstractMatrix`, of size: $(size(g_full[1])), received $(typeof(g_final)).")
            )
        end
    end

    for (idx, ir) in enumerate(iter_range)
        push!(u_tp, get_u(ekp, ir)) #N_parameters x N_ens
        push!(g_tp, g_full[idx]) #N_data x N_ens
    end
    u_tp = reduce(hcat, u_tp) # N_parameters x (N_ek_it x N_ensemble)]
    g_tp = reduce(hcat, g_tp) # N_data x (N_ek_it x N_ensemble)

    training_points = PairedDataContainer(u_tp, g_tp, data_are_columns = true)

    return training_points
end

```

odunbar:
Co-authored by @ArneBouillon 

## Purpose 

- Derived from the LIS work of #376 
- Closes #370

## Content

 - Migration of the LIS from #376 under the following simplifications/changes:
   - [change] `encoder_kwargs` for structure vectors now keep `samples_in` and `samples_out` in their respective distributions, and also contain the algorithm time values `dt` corresponding to the distributions
   - [simplification] we no longer provide the `dim_criterion`, only a threshold for doing dimension reduction
   - [change] the user provides `iters` (Vector{Int}) not `alphas` (Vector{FT}) to indicate which `dt` values they would like to use for dimension reduction. A message reports the `dt` of the requested iterations 
   - [change] if multiple `iters` are requested, by default we compute the LIS for each `iter`, then use a trapezoidal rule to create the final subspace.
   - [simplified] Reduced the complexity of the DimReduction example
   - [change] added a shift by the data mean 
   - [change] Use a more robust information measure `f(k) = log(1+k)^2` and truncate based on `cumsum(f(k))/sum(f(k)) > eps`. Leads to cutoff `eps` that is not "0.99999"
- Extended `encoder_kwargs_from(eki,prior; g_final=...)` to obtain all kwargs for encoding a typical problem. Additionally, we provide a convenience where the user can provide the "g" ensemble for a final evaluation of "u" in from the eki
- Recreated the `DimensionReduction` example for a linear problem to assess performance

## MISC
- Refactor the boundary between public/private encode/decode functions.
    - API (to Utilities): `encode_data`, `encode_structure_matrix`, `encode_with_schedule` applied to `encoder_schedule`. NOT applicable to processors.
    - Private: `_encode_data` `_encode_structure_matrix` only applicable to exact processors etc. 
    - API (to Emulators): `encode_data` and `encode_structure_matrix` (and decode counterparts) allow application to `emulator` in place of `encoder_schedule`
- add small check for error message 
- created #418 as I have not included a docs page for the new example.
- One concern I have: The `noise_injector` in the forward map wrapper can induce instability in the MCMC algorithm, so we might consider changing the default here. (this is not for inflating the posterior, but for evaluating the likelihood when providing reduced parameters to the forward map. This may really be a symptom from the need to resample with MC over several forward map evaluations when injecting the noise)
- Closes #419 Small (breaking) bugfix spotted when resolving codecov 

## Example `DimensionReduction`
We just try the DR methods (using the forward map wrapper in place of an emulator) for a linear example in 100D->100D: The following is output
```
┌ Info: (reduced_dim, error) of posterior mean to whitened "reference" solution,
└  when using the truncation ({for PCA}, {for LI})
5×6 DataFrame
 Row │ reference        PCA-in-out      PCA(mat)-in-out  LI-in(1:1)-out(1:1)  LI-in(1:5)-out(1:1)  truncation       
     │ Tuple…           Tuple…          Tuple…           Tuple…               Tuple…               Tuple…           
─────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ (100, 0.277785)  (41, 0.713683)  (55, 0.306185)   (50, 0.303924)       (32, 0.29261)        (0.995, 0.99995)
   2 │ (100, 0.319642)  (32, 0.798957)  (38, 0.320583)   (34, 0.312999)       (23, 0.339262)       (0.99, 0.9995)
   3 │ (100, 0.29158)   (23, 0.825192)  (23, 0.344845)   (21, 0.323374)       (17, 0.354786)       (0.98, 0.995)
   4 │ (100, 0.309952)  (12, 0.815737)  (11, 0.484712)   (14, 0.390514)       (12, 0.433838)       (0.95, 0.97)
   5 │ (100, 0.301116)  (7, 0.781519)   (6, 0.542278)    (9, 0.520454)        (8, 0.459082)        (0.9, 0.9)
```
**We see that as you increase truncation, our new LI methods reduce errors comparably, or more more slowly, than PCA alone. We also see that using only PCA from EKP samples (over using the prior and obs. matrices here) lead to poor errors**


docs/src/data_processing.md
```diff
@@ -12,37 +12,72 @@ The framework works as follows
...
+```
+emulator = Emulator(
+    machine_learning_tool,       
+    input_output_pairs;
+    encoder_schedule = encoder_schedule,
+    encoder_kwargs = encoder_kwargs_from(ekp, prior), # kwargs from "Calibrate" objects
+)
+```
+
+## Recommended: get `encoder_kwargs` from and `EnsembleKalmanProcess` and `Prior`
```
ArneBouillon: Something seems wrong in this sentence

src/Utilities/likelihood_informed.jl
```diff
@@ -0,0 +1,371 @@
...
+    grad_type::Symbol
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Constructs the `LikelihoodInformed` struct. Keywords:
+- `retain_info`: the method will attempt to limit the KL divergence of the true posterior from the reduced posterior to a value proportional to (1 - `retain_info`). Choose `retain_info` close to 1 to get a good approximation in a large subspace, and reduce it to get a worse approximation in a smaller subspace.
+- `iters`[=[1]]: the likelihood-informed data processor requires samples from the distribution `∝ π_prior(x) π_likelihood(y | x)^α` with `α ∈ [0, 1]`. Here, `iter` indicates the structure vector iterations to use, as sampled from these distributions. For how to pass in these samples, see the `use_data_as_samples` parameter.
+- `grad_type`[=:localsl]: how the gradient of the forward model at the samples will be approximated. Choose from `:linreg` (global linear regression) and `:localsl` (localized statistical linearization; see [Wacker, 2025]).
```
ArneBouillon: For now, I would default this to :linreg, just for robustness

src/Utilities/likelihood_informed.jl
```diff
@@ -0,0 +1,371 @@
...
+        else
+            @warn "Consider using decorrelate_structure_mat to gain obs_noise_cov = I before calling likelihood_informed"
+            false
+        end
+        noise_cov_inv = inv(obs_noise_cov)
+
+        li.apply_to = apply_to
+
+        for (it, α) in zip(iters, alphas[iters]) # take the iterations from alpha
+            #(NB! "it" may not be 1:end)
```
ArneBouillon: Do you mean `iters`? (In the comment)

ArneBouillon (approved):
It looks good and I generally like the design, so let's get this merged!

When the paper is fully finalized, I might update the code with some things I improved while fine-tuning the experiments (e.g., performance improvements)

src/Utilities/likelihood_informed.jl
odunbar: the tool actually does, i think the docstring was inconsistent, resolved

src/Utilities/likelihood_informed.jl
odunbar: yes

PR #413, opened 2026-04-21
New README example for CES

odunbar:
## Purpose 

See README [here](https://github.com/CliMA/CalibrateEmulateSample.jl/tree/orad/docs-enc-dec)

## Content

- README with example inspired from EKP readme.

##MISC
- Slightly increase test tol that causes intermittant failure

## Results from "copy-paste" of snippets into fresh Julia repl:
Calibrate:
 
Emulate:
 
Sample:

----


odunbar:
As Only change is adding readme, will merge on successful tests and reproduction of README under copy-paste.

PR #438, opened 2026-06-25
helper and docs for minibatching

### LINKED ISSUE #436: Minibatching compatability with CES, internalize flattening operations
## Issue

When using EnsembleKalmanProcesses.jl with several observations $$(y_1,\dots,y_N)$$ one might use minibatching to reduce costs. This creates $$(B_1,\dots,B_n)$$ for $$n \leq N$$ and subdivides the forward map $$G = [G_{B_1}(\theta),\dots,G_{B_n}(\theta)] $$, then performs the $$k$$-th  EKP_update based on $$G_{B_k}$$ only. While this can be used as a time-saving/complexity reduction assumption to help use EKP optimize with expensive models, it generates pairs of the form:

$$
\Big( \big(\theta_{1},  G_{B_1}(\theta_1) \big), ..., \big(\theta_{k},  G_{B_k}(\theta_k) \big), \Big)
$$
therefore we do not have pairs $$\theta_i, G(\theta_i)$$

**CES will currently handle the limited setting of multiple minibatches but only if they are size 1**

There is a small modifier to handle most cases we need though. By assuming that the minibatches are from different evaluations of a stochastic simulator, This is the most likely use case for most users data, as often the different data $y_i$ come from this underlying assumption. This would mean that really our data is "y", and not a stacked object. Then by construction of $$B_i$$, this is just sampling from $$y_1,\dots,y_N$$, so just multiple copies of the "y". We in theory have to just break down the batches into their individual components $$B_1 = \{i_1,i_2,i_3...\}$$ that each look like "y", giving pairs :

$$ 
\Big( \big(\theta_{1},  G_{i_1}(\theta_1) \big), ..., \big(\theta_{K},  G_{i_K}(\theta_k) \big), ...\Big)
$$

for $$G_{i_1}$$ being the forward map component fitting only to $$y_{i_1}$$. We can then train an emulator  that treats all the $$G_{i}$$ as random realizations of and underlying "$$G$$", again, this is true for most users.

In this case really we just require a flattening operation to be done by the user as follows:
- get_u and get_g eki iterations
- get the minibatch epochs.
- for each minibatch (size e.g. 3), duplicate any parameter "u" 3 times, and split the output "g" into it's un-batched form"g=[g_1,g_2,g_3]" , then flatten the list `(u,g_1), (u,g_2),(u, g_3)` 
- create the training/test data from here and continue with "CES"
- Also modify encoder_kwargs to be consistent.

Note this is nice though, as it still allows for 
- batches of different sizes (i.e. different numbers of y_i's) 
- batches to overlap (i.e. a batches [1,2,3] and [2,3,4])

## Improvement

As stated in the most broad sense, most users do (and all users _should_ ) use batching in this way where we assume batches just stack observational "atoms" that are all statistically the same. In this way we can modify internal operations and default behaviour to capture this:
- the minibatcher should internalize this flattening so the user does not need to perform it externally
- We also directly extract encoder_kwargs (that may include e.g. output samples) from EKI, and these do not flatten the batches, this sould be updated to do so by default

odunbar:
## Purpose 

- Closes #436 
See docs [here](https://clima.github.io/CalibrateEmulateSample.jl/previews/PR438/emulate/#Typical-construction-from-Lorenz_example.jl)

## Content

- add flattening helper, and use it in `get_training_points` and `encoder_kwargs_from`. The helper takes pairs `(u,g_batched)` with for example with `g_batched` of size `3*dim_y x n_samples` and converts it to  `( (u,g_1), (u,g_2), (u,g_3))` where `g_i` is of size `dim_y x n_samples`. It allows for different sized minibatches
- this helper is now used int `get_training_points(...)` and `encoder_kwargs_from(...)` 
- add unit tests for the helper
- add docs to explain the behaviour with minibatching. Included a note about increased sensitivity to noise distribution when one has "same-input-different-output" data

## MISC
- Created issue #440 to resolve the `@ref` in docs

----

### DIFF
--- a/docs/src/calibrate.md
+++ b/docs/src/calibrate.md
@@ -19,3 +19,22 @@ Documentation on how to construct an EnsembleKalmanProcess from the computer mod
 
 One draw of our approach is that it does not require the forward map to be written in Julia. To aid construction of such a workflow, EnsembleKalmanProcesses.jl provides a documented example of a BASH workflow for the [sinusoid problem](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/examples/sinusoid_example_toml/), with source code [here](https://github.com/CliMA/EnsembleKalmanProcesses.jl/tree/main/examples/SinusoidInterface). The forward map interacts with the calibration tools (EKP) only though TOML file reading an writing, and thus can be written in any language; for example, to be used with [slurm HPC scripts](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/examples/ClimateMachine_example/), with source code [here](https://github.com/CliMA/EnsembleKalmanProcesses.jl/tree/main/examples/ClimateMachine).
 
+## Extracting training data after calibration
+
+Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
+
+**`get_training_points(ekp, n)`** collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
+
+```julia
+using CalibrateEmulateSample.Utilities
+input_output_pairs = get_training_points(ekp, 5)              # first 5 iterations
+input_output_pairs = get_training_points(ekp, 1:2:9)          # odd iterations 1,3,5,7,9
+input_output_pairs = get_training_points(ekp, 5; g_final = G(get_ϕ_final(prior, ekp)))
+```
+
+**`encoder_kwargs_from(ekp, prior)`** collects additional information from the calibration — the prior covariance, the observational noise covariance, and the sequence of parameter and output sample distributions across EKP iterations — into a named tuple that can be passed directly as the `encoder_kwargs` argument of the `Emulator`. This enables data-adaptive preprocessing such as likelihood-informed subspace reduction.
+
+```julia
+encoder_kwargs = encoder_kwargs_from(ekp, prior)
+```
+
--- a/docs/src/calibrate.md
+++ b/docs/src/calibrate.md
@@ -23,13 +23,13 @@ One draw of our approach is that it does not require the forward map to be writt
 
 Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
 
-[`get_training_points(ekp, n; g_final = nothing)`](@ref CalibrateEmulateSample.Utilities.get_training_points) collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
+[`get_training_points(ekp, n; g_final = nothing)`](@ref CalibrateEmulateSample.Utilities.get_training_points) collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://clima.github.io/EnsembleKalmanProcesses.jl/dev/internal_data_representation/) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
 
 ```julia
 using CalibrateEmulateSample.Utilities
 input_output_pairs = get_training_points(ekp, 5)              # first 5 iterations
-input_output_pairs = get_training_points(ekp, 1:2:9)          # odd iterations 1,3,5,7,9
-input_output_pairs = get_training_points(ekp, 5; g_final = G(get_ϕ_final(prior, ekp)))
+input_output_pairs = get_training_points(ekp, 1:2:9)          # odd iterations 
+input_output_pairs = get_training_points(ekp, 5; g_final = G.(get_ϕ_final(prior, ekp)))  # an additional iteration pairing (u_final, g_final) 
 ```
 
 [`encoder_kwargs_from(ekp, prior)`](@ref CalibrateEmulateSample.Utilities.encoder_kwargs_from) collects additional information from the calibration — the prior covariance, the observational noise covariance, and the sequence of parameter and output sample distributions across EKP iterations — into a named tuple that can be passed directly as the `encoder_kwargs` argument of the `Emulator`. This enables data-adaptive preprocessing such as likelihood-informed subspace reduction.
--- a/docs/src/calibrate.md
+++ b/docs/src/calibrate.md
@@ -23,7 +23,7 @@ One draw of our approach is that it does not require the forward map to be writt
 
 Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
 
-**`get_training_points(ekp, n)`** collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
+**`get_training_points(ekp, n; g_final = nothing)`** collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
 
 ```julia
 using CalibrateEmulateSample.Utilities
--- a/docs/src/calibrate.md
+++ b/docs/src/calibrate.md
@@ -23,7 +23,7 @@ One draw of our approach is that it does not require the forward map to be writt
 
 Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
 
-[`get_training_points(ekp, n; g_final = nothing)`](@ref Utilities.get_training_points) collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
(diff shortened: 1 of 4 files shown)


docs/src/calibrate.md
```diff
@@ -19,3 +19,22 @@ Documentation on how to construct an EnsembleKalmanProcess from the computer mod
...
 
+## Extracting training data after calibration
+
+Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
+
+**`get_training_points(ekp, n)`** collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
```
ph-kev: It would also be a good idea to add an `@ref` for the function.

```suggestion
**`get_training_points(ekp, n; g_final = nothing)`** collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
```

docs/src/emulate.md
```diff
@@ -11,6 +11,14 @@ First, obtain data in a `PairedDataContainer`, for example, get this from an `En
 using CalibrateEmulateSample.Utilities
 input_output_pairs = Utilities.get_training_points(ekpobj, 5) # use first 5 iterations as data
 ```
+
+!!! note "Minibatched calibration"
+    When the `EnsembleKalmanProcess` is built with an `ObservationSeries` and a minibatcher, each iteration uses a batch ``B_1, \ldots, B_n`` (``n \leq N``): a disjoint partition of statistically similar observations ``y_1, \ldots, y_N \sim \rho``. The batching can aid the calibration stage but does not effect the emulate-sample stages.
```
ph-kev: Is it meaning of `N` obvious here? I am not sure what it is referring to.

```suggestion
    When the `EnsembleKalmanProcess` is built with an `ObservationSeries` and a minibatcher, each iteration uses a batch ``B_1, \ldots, B_n`` (``n \leq N``): a disjoint partition of statistically similar observations ``y_1, \ldots, y_N \sim \rho``. The batching can aid the calibration stage but does not affect the emulate-sample stages.
```

docs/src/emulate.md
```diff
@@ -11,6 +11,14 @@ First, obtain data in a `PairedDataContainer`, for example, get this from an `En
...
+
+    Internally `get_training_points` and `encoder_kwargs_from` reduce outputs/noise from the `EnsembleKalmanProcess` to thei per-observation components, thus the emulator is trained on ``(\theta, \mathcal{G}(\theta))`` pairs representative of the distribution ``\rho``. In the Sampling stage the full observation set ``y_1, \ldots, y_N`` is used and batch structure is ignored.
```
ph-kev: You can also use `@ref` here too.

docs/src/emulate.md
```diff
@@ -11,6 +11,14 @@ First, obtain data in a `PairedDataContainer`, for example, get this from an `En
...
+
+    Warning: using training data with several `(identical-input)->(different-output)` pairs, will induce more sensitivity on the `obs_noise_cov`. It is recommended to ensure it is well estimated (or well-regularized with white noise) to help training; one can also use the `noise_learn=true` flag but this is known to cause slow-downs in training.
```
ph-kev, line 20: You can use Documenter.jl admonition for this. The syntax for the suggestion is possibly off. 

```suggestion
!!! warning
    using training data with several `(identical-input)->(different-output)` pairs, will induce more sensitivity on the `obs_noise_cov`. It is recommended to ensure it is well estimated (or well-regularized with white noise) to help training; one can also use the `noise_learn=true` flag but this is known to cause slow-downs in training.
```

src/Utilities.jl
```diff
@@ -162,6 +142,121 @@ end
...
+`batch_size` blocks of `gdim` rows and replicates the corresponding parameter ensemble
+so that each batch element becomes an independent `(u, g)` training pair. For
+non-minibatched runs (`batch_size = 1`) the function is a no-op.
+
+# Arguments
+
+- `ekp`: `EnsembleKalmanProcess` holding the parameter ensembles and forward-model outputs.
+- `iter_range`: iterations to include — an integer `n` is interpreted as `1:n`
+  (clamped to the number of stored outputs); a vector selects specific iterations,
+  which must be a subset of `1:length(get_g(ekp))`.
```
ph-kev, line 179: Does order matters for the iterations? For example, can I pass `[5,4,2,1]`?

src/Utilities.jl
```diff
@@ -162,6 +142,121 @@ end
...
+                    "iter_range contains iteration $ir, but EKP only stores $g_len " *
+                    "outputs. iter_range must be a subset of 1:length(get_g(ekp)).",
+                ),
+            )
+            gg = final_samples_out
+        else
+            gg = get_g(ekp, ir)
+        end
+        uu = get_u(ekp, ir)
+        batch = get_minibatch(os, ir)
```
braghiere, line 237: Can you confirm get_minibatch(os, ir) returns the batch that actually produced get_g(ekp, ir), including after the minibatcher wraps around?

src/Utilities.jl
```diff
@@ -162,6 +142,121 @@ end
...
+    os = get_observation_series(ekp)
+    g_len = length(get_g(ekp))
+    alg_time = include_dt ? get_algorithm_time(ekp) : nothing
+
+    u_out = []
+    g_out = []
+    dt_out = include_dt ? Float64[] : nothing
+
+    # if a final output is provided, treat it as one extra iteration beyond g_len
+    full_range = isnothing(final_samples_out) ? iter_range : [collect(iter_range); g_len + 1]
```
braghiere, line 221: The final_samples_out contract (stack as [g_b1; g_b2; …] matching the next cycle's batches) is easy to get wrong, and it's only checked for divisibility. a wrong order silently mis-pairs noise. I avoided this by evaluating the final ensemble on the full obs vector. Could it take the unbatched output and split deterministically, or at least validate the ordering?

braghiere (commented):
Nice! I hit exactly this (get_training_points couldn't stack different-sized minibatch outputs) calibrating the land model, so glad it's fixed upstream. The noise note is a great call. A few things below.

docs/src/emulate.md
odunbar: the number of batches/samples. I have highlighted

src/Utilities.jl
odunbar: I don't think so, the training points are flattened.

docs/src/emulate.md
odunbar: This is supposed to be inside the notes block though. it doesn't need to be a warning block - it's just an addendum, i don;t think it should be its own block

src/Utilities.jl
odunbar: Should do! This is the tooltip 
```julia
help?> get_minibatch

get_minibatch(
      os::ObservationSeries,
      it_or_mbi::Union{Nothing, Int, Dict}
  ) -> Any
  
get the minibatch for a given minibatch index `Dict("epoch"=> x, "minibatch" => y)`, or iteration
`Int`. If `nothing` is provided as an iteration then the current minibatch is returned
```
so get_minibatch(os, 7) gets the minibatch for the observation used from the 7th iteration.

RE your wrapping around question. Yes: if you have 4 batches in an epoch, then iteration "7" will be converted into epoch 2 batch 3, for example

src/Utilities.jl
odunbar: I am not sure i follow. Passing in unbatched `g_batch`  still doesn't help because if the minibatch are for the observations `[15,7,2,34]` then how are you ensuring that you pass in the four g_batch items in the correct order? Or do you mean that you are running the model to produce the entirity of all minibatches e.g. `[1,...,100]` and then extracting the 4 you want? Because this kind of sidesteps the entire minibatching procedure.

We also can't check ordering, as `g_final` is always a matrix, so there is no way to validate generally what order it is. 

However i would say it isn't necessarily "easy to get wrong" as you just run "the next iteration" using the gotten minibatch from EKP. As the user has already done this for the iterations in EKP it should just be akin to calling it again (without the `update_ensemble!` call). to ensure folks are aware of the syntax though I have added this into the docstring:

```
get_training_points(
      ekp::EnsembleKalmanProcesses.EnsembleKalmanProcess{FT, IT, P},
      train_iterations::Union{AbstractVector{IT}, IT} where IT;
      g_final
  ) -> EnsembleKalmanProcesses.DataContainers.PairedDataContainer
  

  Extract and flatten the training data from an EnsembleKalmanProcess into a PairedDataContainer
  suitable for training an Emulator.

  Arguments
  ≡≡≡≡≡≡≡≡≡

    •  ekp: EnsembleKalmanProcess holding the parameter ensemble and forward-model outputs.
    •  train_iterations: integer n (uses iterations 1:n) or an index vector such as 3:2:9.
    •  g_final (keyword, default nothing): optional AbstractMatrix of forward-model outputs for
       the final parameter ensemble (not yet stored in ekp), sized as get_g(ekp, 1).

  To ensure consistency in g_final, we recommend that, given the ekp::EnsembleKalmanProcess and
  prior::ParameterDistribution used to generate it, the user evaluates the forward map G on the
  parameters returned by get_ϕ_final:

  params = get_ϕ_final(prior, ekp)
  g_final = reduce(hcat, [G(param) for param in eachcol(params)])

  If ekp is using minibatching, G must be evaluated on the same minibatch that ekp is currently on:

  params = get_ϕ_final(prior, ekp)
  batch_final = get_current_minibatch(ekp)
  g_final = reduce(hcat, [G(param, batch_final) for param in eachcol(params)])

  See EnsembleKalmanProcesses.Observations for more on minibatching and ObservationSeries.

```

docs/src/emulate.md
odunbar: I just tried adding  @ref in the calibrate file, and it would not resolve in the deployment. As this is annoying, and we have never used @ref to methods (just between docs sections). I think I may do a docs-wide sweep to address this in future PR. I'll create an issue instead.

odunbar: #440

docs/src/calibrate.md
```diff
@@ -19,3 +19,22 @@ Documentation on how to construct an EnsembleKalmanProcess from the computer mod
...
 
+## Extracting training data after calibration
+
+Two helper functions from `CalibrateEmulateSample.Utilities` simplify passing the calibration results to the emulator.
+
+[`get_training_points(ekp, n; g_final = nothing)`](@ref CalibrateEmulateSample.Utilities.get_training_points) collects the stored parameter ensembles and forward-model outputs into a [`PairedDataContainer`](https://github.com/CliMA/EnsembleKalmanProcesses.jl/blob/main/src/DataContainers.jl) ready for emulator training. The argument `n` can be an integer (which uses iterations `1:n`) or an index vector.  An optional keyword `g_final` accepts the forward-model outputs at the final, not-yet-stored parameter ensemble.
```
ph-kev: I find it strange that the reference for `PairedDataContainer` is a link to a source file instead of a page of the public facing documentation of EnsembleKalmanProcesses.jl.

ph-kev (approved):

docs/src/calibrate.md
odunbar: I'll fix this one, and add this to the linking issue mentioned above.

PR #448, opened 2026-07-22
MCMC proposal bug-fixes and small refactor

### LINKED ISSUE #447: [audit 2026-07-09] overly Symmetrised proposals in MCMC
## Issue 

Not all proposals are symmetric, for example the pCN proposal here:
https://github.com/CliMA/CalibrateEmulateSample.jl/blob/26d74557f61001f221d61527dc12663e908847fd/src/MarkovChainMonteCarlo.jl#L343-L354

has a transition density that is not-symmetric i.e. q(u,u') != q(u',u). However the `logratio_proposal_density` that should do "q(u,u') - q(u',u)"  directly when non-symmetric. Instead, all methods currently default to a `_get_proposal` call that was implemented for RWM, and a transition that assumes a symmetric update.
https://github.com/CliMA/CalibrateEmulateSample.jl/blob/26d74557f61001f221d61527dc12663e908847fd/src/MarkovChainMonteCarlo.jl#L164-L176

This applies to pCN and Barker and the effect e.g. for pCN is that there is a doubled prior contribution.

## Possible solution:
I think the bug was from some copied boilerplate (due to a misconception of the messy AdvancedMH interface we had to use). It is a relatively straightforward change. One simply writes the transition density for each code, (not based on the same `_get_proposal`) and then  fully write out the ratio as "reverse - forward". In the symmetric case this cancels but otherwise it will not.
```julia
# Each sampler implements this
function log_transition_density(sampler::MHS, θ_from, θ_to) where {MHS <: AdvancedMH.MHSampler}      
    ...
end   

# Forward and reverse called by all samplers
function AdvancedMH.logratio_proposal_density(sampler::MHS, transition_prev, candidate) where {MHS   <: AdvancedMH.MHSampler}
       log_q_forward = log_transition_density(sampler, transition_prev.params, candidate)  #    q(cand| prev)       
       log_q_reverse = log_transition_density(sampler, candidate, transition_prev.params)  # q(prev | cand)       
       return log_q_reverse - log_q_forward  
end
```

I think furthermore, for many parametric types, we could implement the Markov Kernel directly, and then the transition becomes the `logpdf(kernel)`  while the `propose()` method just becomes a "rand!(kernel)", reducing the multiplicity. However this might not apply to more generic kernels such as `Barker`...

odunbar:
## Purpose 

- Closes #447 

## Content

- Fix the issue where proposals default (via copied boilerplate) to RWM log-ratio framework that is incompatible with asymmetric densities.
- This also requires a bug-fix in the Barker proposal (which assumes decorrelated sampling, so we use the cholesky factors to whiten)
-  In their place, we define the proper log ratio by using a correct transition density "q(u|u') - q(u'|u)" 
- We have a small refactor, to define the Markov kernel, as a Struct. This allows for kernels that can be represented by Distributions.jl, we make a single location to define the kernel, then rand(kernel) produces a proposal, and logpdf(kernel) produces the transition log-density. For kernels that aren't representable by Distributions.jl (i.e. Barker) we store the relevant parameters in the kernel, and use similar interface for readability via methods  `_barker_rand(kernel)` and `_barker_logpdf(kernel)` with the analytical formulas inside.
- We have unit tests that will catch future bugs in the proposals. We ensure that some analytic posteriors are recovered well
- Docstrings and docs API

## MISC
- we also improved the stepsize criteria to a bisection method rather than a doubling approach
- Add LogExpFunctions for more stable log(1/(1+exp(x))) functions

----


odunbar:
Noone to review MCMC pipelines at the moment, will defer to the new analytic tests showing that failing tests now pass. I have also verified the math of the updates with the asymmetric proposals.

